PR #26, opened 2023-04-09
use {pacta.data.scraping}

### LINKED ISSUE #25: utilize functions directly from {pacta.data.scraping}
[Data scraping functions have been extracted](https://github.com/RMI-PACTA/pacta.data.preparation/issues/282) from [{pacta.data.preparation}](https://github.com/RMI-PACTA/pacta.data.preparation) into [{pacta.data.scraping}](https://github.com/RMI-PACTA/pacta.data.scraping) (and will eventually be removed from {pacta.data.preparation}).

We should depend on {pacta.data.scraping} here and utilize its functions directly.

cjyetman:
closes #25 
depends on https://github.com/RMI-PACTA/pacta.data.scraping/issues/3

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -288,7 +288,7 @@ log_info("Preparing scenario data... ")
 
 scenario_regions <- readr::read_csv(scenario_regions_path, na = "", show_col_types = FALSE)
 
-index_regions <- pacta.data.scraping::get_index_regions()
+index_regions <- pacta.data.preparation::index_regions
 
 factset_issue_code_bridge <- pacta.data.preparation::factset_issue_code_bridge %>%
   select(issue_type_code, asset_type) %>%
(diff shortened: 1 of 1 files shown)


AlexAxthelm (approved):

cjyetman:
I'm putting this in draft mode because the scraper functions are a bit flaky. That's nothing new, they always are/have been, but for instance we're now running `get_index_regions()` on the fly here rather than pre-generating it in a different process as we have done before.

Need to think about how to do this in a way that's not going to cause a lot of frustration when just trying to run a new data.prep process (especially while the data.prep process takes ~6 hours, so it's not something you want to start and stop multiple times if not necessary).

jdhoffa (changes requested):
See comment above

cjyetman:
> As per our call today, I would suggest splitting this into two PRs. The first of which is a pure refactor, that implements the exact same functionality as before, but using `pacta.data.scraping`.
> 
> In the second PR, we can explore using `get_index_regions()` as opposed to the internal dataset.
> 
> This way we are able to easily revert the commit, if we find that the scraping functionality is unstable.

done

jdhoffa (approved):
lgtm

PR #29, opened 2023-04-11
build index regions on the fly with new `pacta.data.scraping::get_index_regions()`

cjyetman:


jdhoffa (approved):
lgtm but should be merged after #26

PR #30, opened 2023-04-11
now that functions are exported, no need to namespace them

cjyetman:
depends on https://github.com/RMI-PACTA/pacta.data.preparation/pull/292
related https://github.com/RMI-PACTA/pacta.data.preparation/issues/275

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -290,7 +290,8 @@ scenario_regions <- readr::read_csv(scenario_regions_path, na = "", show_col_typ
 
 index_regions <- pacta.data.scraping::get_index_regions()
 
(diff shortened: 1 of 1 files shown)


cjyetman:
ok... I'll edit this to only use `<package>::<function>`, but not `<package>:::<function>` for unexpected functions

jdhoffa:
Yes absolutely, we should avoid using internal functions.

cjyetman:
it makes some lines really long and functions names absurdly long, but it is more clear

jdhoffa:
> it makes some lines really long and functions names absurdly long, but it is clear

yeah fair enough, I'm generally fine with absurdly long lines if they are explicit. Future me will always appreciate something explaining things to me like I'm 5 over something that is pretty but hard to understand.

PR #32, opened 2023-04-12
include text of NEWS.md from relevant PACTA packages in manifest para…

cjyetman:
…meters

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -782,6 +782,19 @@ ent_entity_affiliates_last_update <-
   pull(ent_entity_affiliates_last_update) %>%
   unique()
 
+# include PACTA packages NEWS.md test in the parameters to export
+pacta_packages <- c("pacta.data.preparation", "pacta.scenario.preparation")
+package_news <-
+  vapply(
+    X = pacta_packages,
+    FUN = function(pkg) {
+      pkg_news_file <- system.file("NEWS.md", package = pkg)
+      list(paste0(readLines(pkg_news_file), collapse = "\n"))
+    },
+    FUN.VALUE = list(1),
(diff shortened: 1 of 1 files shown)


run_pacta_data_preparation.R
```diff
@@ -782,6 +782,19 @@ ent_entity_affiliates_last_update <-
   pull(ent_entity_affiliates_last_update) %>%
   unique()
 
+# include PACTA packages NEWS.md test in the parameters to export
+pacta_packages <- c("pacta.data.preparation", "pacta.scenario.preparation")
```
jdhoffa, line 786: WET: These packages are defined again here:
https://github.com/RMI-PACTA/workflow.data.preparation/blob/331f6abf1bbe1ae990960c87a09b8de6a2e63deb/run_pacta_data_preparation.R#L855
and here:
https://github.com/RMI-PACTA/workflow.data.preparation/blob/331f6abf1bbe1ae990960c87a09b8de6a2e63deb/run_pacta_data_preparation.R#L861

for essentially the same purpose. 

Would pose less maintenance burden to define the pacta_packages you want to include NEWS information from once, and use the same argument for both copying over the files, and writing the manifest.

jdhoffa (changes requested):

run_pacta_data_preparation.R
cjyetman: agreed... I didn't wanna change anything contextually separate here... let's leverage the new `pacta_packages` once this PR is merged in a new PR that cleans up the `NEWS.md` copying bit

cjyetman: #35

jdhoffa (approved):
Given #35, this LGTM

PR #34, opened 2023-04-12
Add `tdm_years` concept to data preparation

### LINKED ISSUE #31: Calculate t0 + 5 and t0 + 10 values to be used in TDM calculation
The TDM metric was conceived in 2020. At that point, it was convenient to calculate, as t+5 and t+10 had very round numbers (2025 and 2030). 

Now, we would like to continue calculating the TDM metric, however, for 2021Q4 data, this requires calculating scenario targets for t+5 = 2026 and t+10 = 2031. 

This would require making some adjustment to values in the `config.yml` here:
https://github.com/RMI-PACTA/workflow.data.preparation/blob/331f6abf1bbe1ae990960c87a09b8de6a2e63deb/config.yml#L40-L41

The `t+5` case is already handled, however the `t+10` case should be included, perhaps by adding `2031` to the `adittional_year` config. Perhaps it is safer/ more future proof to just add a `tdm_years` configuration parameter, that way we know what we are keeping track of (and can remove it easily if we decide to remove the metric).

We may also consider double-checking WHY we need, for example, the year `2040`.

This is a whole big headache lol.

See further discussion here: https://dev.azure.com/RMI-PACTA/2DegreesInvesting/_workitems/edit/6197

Relates to https://github.com/RMI-PACTA/pacta.portfolio.analysis/pull/221

jdhoffa:
This PR:
* Adds a new parameter, `tdm_delta_years`, to the `config.yml`, which indicates any extra years that scenario targets need be calculated for, to be used in the TDM. These years are "deltas" meaning a duration, as measured from the start year (rather than a fixed value)
* Adds these new years to the `run_pacta_data_preparation` script

What this PR does NOT do: Test if those years are even present in any of the scenario data!!

Relates to https://github.com/RMI-PACTA/pacta.portfolio.analysis/issues/219

Closes #31

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -15,6 +15,7 @@ default:
   indices_timestamp: "20211231"
   market_share_target_reference_year: 2021
   time_horizon: 5
+  tdm_delta_years: [5, 10]
   additional_year: [2030, 2040]
   scenario_sources_list: ["ETP2020", "GECO2021", "IPR2021", "ISF2021", "WEO2021"]
   sector_list: ["Automotive", "Power", "Fossil Fuels", "Oil&Gas", "Coal"]
@@ -38,6 +39,7 @@ default:
   indices_timestamp: "20211231"
   market_share_target_reference_year: 2021
   time_horizon: 5
+  tdm_delta_years: [5, 10]
   additional_year: [2030, 2040]
   scenario_sources_list: ["ETP2020", "GECO2021", "IPR2021", "ISF2021", "WEO2021"]
   sector_list: ["Automotive", "Power", "Fossil Fuels", "Oil&Gas", "Coal"]
@@ -70,6 +72,7 @@ default:
   indices_timestamp: "20220630"
   market_share_target_reference_year: 2022
   time_horizon: 5
+  tdm_delta_years: [5, 10]
   additional_year: [2030, 2040]
   scenario_sources_list: ["ETP2020", "GECO2021", "IPR2021", "ISF2021", "WEO2021"]
   sector_list: ["Automotive", "Power", "Fossil Fuels", "Oil&Gas", "Coal"]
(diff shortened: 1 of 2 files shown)


run_pacta_data_preparation.R
```diff
@@ -290,7 +291,7 @@ scenario_regions <- readr::read_csv(scenario_regions_path, na = "", show_col_typ
 
 index_regions <- pacta.data.scraping::get_index_regions()
 
-factset_issue_code_bridge <- 
+factset_issue_code_bridge <-
```
jdhoffa, line 294: Sorry I think this whitespace got removed automatically on save.

config.yml
```diff
@@ -38,6 +39,7 @@ default:
   indices_timestamp: "20211231"
   market_share_target_reference_year: 2021
   time_horizon: 5
+  tdm_delta_years: [5, 10]
```
cjyetman, line 42: I think maybe it is conceptually better to allow the `2021Q4` and `2022Q4` configs to inherit this parameter from `default`, because otherwise it implies that this value is specific to this config

(any parameters in `default` but not in another named config are automatically inherited, e.g. `update_factset`)

jdhoffa: Ok, I was just following suit to other parameters (for example `time_horizon` is also likely to be identical between 2021Q4 and 2022Q4 runs)

cjyetman: #36

cjyetman: yes, I believe any values found in `default` that are not found in another config are inherited, e.g. `update_factset`

that's why this works when the `2021Q4` config is used
https://github.com/RMI-PACTA/workflow.data.preparation/blob/cf842c281b00df695f5a5433cb6ffed5cecab330/run_pacta_data_preparation.R#L24-L40

---

https://rstudio.github.io/config/articles/inheritance.html
> Other configurations automatically inherit all default values so need only define values specialized for that configuration.

cjyetman (approved):

Issue #27, opened 2023-04-09
remove indices/benchmark collection and processing


cjyetman:
related https://github.com/RMI-PACTA/pacta.data.preparation/issues/278
related https://github.com/RMI-PACTA/workflow.prepare.pacta.indices/issues/20

As discussed here https://github.com/RMI-PACTA/workflow.data.preparation/issues/2#issuecomment-1501078526, our current strategy is to separate the data.prep and indices.prep processes because the data.prep output *without the benchmark data* is still a useful, complete set of data for pacta.portfolio.analysis and has a legitimate use case. For a use case like transition monitor, that should require as an additional input the prepared indices results, which should be separately prepared by the indices.prep processes.

Therefore, we should remove all of the code here that downloads, processes, and exports the indices data. e.g.

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1b81a59df7a04e634eeaf9c867743f1b107f4bea/run_pacta_data_preparation.R#L41

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1b81a59df7a04e634eeaf9c867743f1b107f4bea/run_pacta_data_preparation.R#L45

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1b81a59df7a04e634eeaf9c867743f1b107f4bea/run_pacta_data_preparation.R#L82-L83

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1b81a59df7a04e634eeaf9c867743f1b107f4bea/run_pacta_data_preparation.R#L105-L122

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1b81a59df7a04e634eeaf9c867743f1b107f4bea/run_pacta_data_preparation.R#L144-L147

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1b81a59df7a04e634eeaf9c867743f1b107f4bea/run_pacta_data_preparation.R#L251-L279

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1b81a59df7a04e634eeaf9c867743f1b107f4bea/run_pacta_data_preparation.R#L365-L377

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1b81a59df7a04e634eeaf9c867743f1b107f4bea/run_pacta_data_preparation.R#L797-L798

[...]

PR #38, opened 2023-04-13
add pacta.data.scraping as non-CRAN dependency

cjyetman:
#27 added pacta.data.scraping as a dependency but didn't add it to the list of non-CRAN package dependencies in the Dockerfile


jdhoffa:
Closing in favour of #39

PR #39, opened 2023-04-13
handle cran and non-cran deps separately

jdhoffa:
Supersedes #38

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,10 +8,9 @@ WORKDIR /workflow.data.preparation
 
 RUN Rscript -e '\
   readRenviron(".env"); \
-  private_pkgs <- c("RMI-PACTA/pacta.scenario.preparation", "RMI-PACTA/pacta.data.preparation"); \
-  script_dep_pkgs <- setdiff(renv::dependencies()$Package, basename(private_pkgs)); \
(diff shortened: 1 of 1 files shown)


cjyetman (approved):

cjyetman:
For more detail...

The `pak::pkg_install("tidyverse/dbplyr@v2.2.1")` line was added to the Dockerfile in 0be326b3a49c217d138c3ee3ac17fdad167a27f1 because at the time pacta.data.preparation had code that broke with the concurrent {dbplyr} CRAN release version, and pacta.data.preparation itself did not specify this version dependency.

That was later changed to `pak::pkg_install("tidyverse/dbplyr")` in #20 because I decided it was better to rely on the newer dev version of {dbplyr} which already had a fix for the problem.

Later, I made pacta.data.preparation specify a version dependency for {dbplyr} (targeting the dev version) in https://github.com/RMI-PACTA/pacta.data.preparation/pull/287

Now that pacta.data.preparation specifies a working version of {dbplyr} in its DESCRIPTION file, we should be able to trust that when pacta.data.preparation is installed, we will get the appropriate/required version of {dbplyr}, which makes doing it explicitly in the Dockerfile unnecessary.

PR #47, opened 2023-04-23
inherit `time_horizon`

### LINKED ISSUE #36: consider which values can/should be inherited from `default` config
Ok, I was just following suit to other parameters (for example `time_horizon` is also likely to be identical between 2021Q4 and 2022Q4 runs)

_Originally posted by @jdhoffa in https://github.com/RMI-PACTA/workflow.data.preparation/pull/34#discussion_r1164026861_

cjyetman:
closes #36 

I think `time_horizon` is the only one we definitely should inherit... @jdhoffa ?

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -35,7 +35,6 @@ default:
   factset_data_timestamp: "2021-12-31"
   pacta_financial_timestamp: "2021Q4"
   market_share_target_reference_year: 2021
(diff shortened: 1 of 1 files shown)


jdhoffa:
Yup agreed.

jdhoffa (approved):
lgtm

PR #49, opened 2023-05-02
Update `tdm_delta_years` to be in bounds of ABCD data

jdhoffa:
Currently, scenario targets are only calculated for years for which we have ABCD data. 

Since we only have data up until 2030, calculating targets for TDM for 2021, 2026 and 2031 won't work.

We also recently revised our decision for choosing an appropriate start year.

Beyond this, it is not in-line with what is described here: https://github.com/RMI-PACTA/pacta.interactive.report/blob/54b3ff30056e398c4dc33890b7194df69910b1f0/general_en_template/rmd/07-IPR.Rmd#L3

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -14,7 +14,7 @@ default:
   pacta_financial_timestamp: "2021Q4"
   market_share_target_reference_year: 2021
   time_horizon: 5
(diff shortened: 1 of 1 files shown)


cjyetman (approved):
easy! LOL

PR #51, opened 2023-05-08
Use new dataset `scenario_source_pacta_geography_bridge` to bridge scenario region names

jdhoffa:
This PR makes use of the newly defined `scenario_source_pacta_geography_bridge` to bridge the scenario-defined region names (e.g. IPR "WORLD") to the PACTA-compatible scenario name (e.g. "Global"). 

It does this using a dataset exported by the pacta.scenario.preparation package. 

Currently the only regions defined and managed are "Global", "OECD" and "NonOECD". 
The only scenarios this applies to are "ETP2020", "GECO2021", "ISF2021", "WEO2021", and "IPR2021".

Prior to adding new scenarios to Transition Monitor, we will need to update this list. 

Depends on https://github.com/RMI-PACTA/pacta.scenario.preparation/pull/91

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -259,6 +259,15 @@ scenario_raw <- readr::read_csv(scenarios_analysis_input_path, show_col_types = 
 
 # filter for relevant scenario data
 scenarios_long <- scenario_raw %>%
+  inner_join(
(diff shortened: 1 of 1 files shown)


cjyetman (approved):
makes sense to me

PR #45, opened 2023-04-14
`config.yml` gains params for `2022Q4` data preparation

jdhoffa:

PR #57, opened 2023-06-21
`config.yml` gains params for `2022Q4_MFM` data preparation

jdhoffa:


config.yml
```diff
@@ -78,3 +78,25 @@ default:
...
+  masterdata_debt_filename: "2023-06-18_AI_RMI Bespoke_Company Data Products_masterdata_debt_2022Q4.csv"
+  ar_company_id__factset_entity_id_filename: "2023-02-15_AI_RMI_Bespoke_Company Data Products_Company_ID_List_2022Q4.csv"
+  imf_quarter_timestamp: "2022-Q4"
+  factset_data_timestamp: "2022-12-31"
+  pacta_financial_timestamp: "2022Q4"
+  market_share_target_reference_year: 2022
+  time_horizon: 5
+  tdm_delta_years: [5, 8]
+  scenario_sources_list: ["GECO2022", "IPR2021", "ISF2021", "WEO2022"]
+  sector_list: ["Automotive", "Power", "Fossil Fuels", "Oil&Gas", "Coal"]
```
Antoine-Lalechere, line 93: probably NIT but why do we include Fossil Fuels here if we already get Oil&Gas and Coal?

Antoine-Lalechere (dismissed):

config.yml
jdhoffa: a very good point!

jdhoffa:
Closing in favour of #45

PR #65, opened 2023-07-24
Update README.md

### LINKED ISSUE #64: Update README
The README has a few things that seem to be out of sync with the current code:
1. `HOST_INPUTS_PATH` and `HOST_OUTPUTS_PATH` env vars from the `.env` file do not appear to be used at all, instead those values are taken from the config file https://github.com/RMI-PACTA/workflow.data.preparation/blob/86881db2d3e8b1141697ec2a82ed5e2bef3f6e09/run_pacta_data_preparation.R#L33-L34
2. `GITHUB_PAT` doesn't seem to be needed/used at all. It seems like the code expects the necessary pkgs to be installed already https://github.com/RMI-PACTA/workflow.data.preparation/blob/86881db2d3e8b1141697ec2a82ed5e2bef3f6e09/run_pacta_data_preparation.R#L4-L6

jdhoffa:
Slight adjustments to `.env` file specifications in README to make intention clear.

Closes #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,12 @@
 
 Running the workflow requires a file `.env` to exist in the root directory, that looks like...
 ``` sh
-HOST_INPUTS_PATH=/users/pacta/inputs
-HOST_OUTPUTS_PATH=/users/pacta/outputs
+HOST_INPUTS_PATH=/PATH/TO/AR_YYYYQQ
+HOST_OUTPUTS_PATH=/PATH/TO/YYYYQQ_pacta_analysis_inputs_YYYY-MM-DD/YYYYQQ
 GITHUB_PAT=ghp_XXXXxxXxXXXxXxxX
 R_DATABASE_USER=xxxx@xxxxx
 R_DATABASE_PASSWORD=xxxXxxXxxxxxxXXxxxxXXXXxx
-R_CONFIG_ACTIVE=2021Q4
+R_CONFIG_ACTIVE=YYYYQQ
 ```
 
 - `HOST_INPUTS_PATH` the local path to where the input files live. This directory will be mounted into the Docker image and it will read and write some files to this directory, so appropriate permissions in this directory will be required. The pacta.data.preparation process requires 3 input files that must exist in this directory and they must have filenames that match those specified in the [config.yml](config.yml) for the specified config:


README.md
```diff
@@ -2,12 +2,12 @@
 
 Running the workflow requires a file `.env` to exist in the root directory, that looks like...
 ``` sh
-HOST_INPUTS_PATH=/users/pacta/inputs
-HOST_OUTPUTS_PATH=/users/pacta/outputs
+HOST_INPUTS_PATH=/PATH/TO/AR_YYYYQQ
+HOST_OUTPUTS_PATH=/PATH/TO/YYYYQQ_pacta_analysis_inputs_YYYY-MM-DD/YYYYQQ
```
cjyetman, line 6: As far as I can tell, these do not have any impact, so I would suggest removing them completely otherwise it just causes confusion.

jdhoffa: https://github.com/RMI-PACTA/workflow.data.preparation/blob/86881db2d3e8b1141697ec2a82ed5e2bef3f6e09/docker-compose.yml#L7-L13

cjyetman (approved):
I'm not opposed to the changes here in general, but they don't address either of the problems mentioned in #64, so I would not close that ticket with this PR

README.md
cjyetman: ok, so this is another situation where what happens with Docker is not the same as what happens when I run it on my machine when I'm testing stuff out

jdhoffa: I mean it's just a specification of volume mounts?

In the container, `run_pacta_data_preparation` will always point to `inputs` and `outputs` (on your local machine it should always point to `inputs` and `outputs`), the only difference is that the `.env` file specifies how to volume mount things into that container

jdhoffa: It's the exact same thing as how we volume mount whatever working directory (from wherever we want) into the `working_dir` folder for `workflow.transition.monitor`. That's just where the code expects the inputs/ outputs to be, but when you spin up the container you can point to wherever you like, and volume mount it there

cjyetman: If you clone this repo to your machine and run `run_pacta_data_preparation.R`, the input and output paths are set here (from the config file, not the .env file)
https://github.com/RMI-PACTA/workflow.data.preparation/blob/86881db2d3e8b1141697ec2a82ed5e2bef3f6e09/run_pacta_data_preparation.R#L33-L34

I just went through this. I read the README, I changed the path in the `.env` file, then I ran it and the paths were wrong, so then I investigated a found that the paths are taken from the config file, so then I edited the paths there instead and everything worked as expected.

jdhoffa: Right, but what I am saying is that, within the Docker container, those are the filepaths that it looks for (`inputs` and `outputs`), it just happens that those directories get volume mounted in. 

And yes, the README has no instructions relevant to running the workflow locally. It expects that you are running it by calling `docker-compose up`

jdhoffa: I think if we want to maintain two expected workflows for running this, we will need to make that explicitly clear in the `README` and explain both workflow simultaneously (e.g. one section on "Running with Docker", and one section on "Running from RStudio"). 

It's unlikely that we will ever end up with instructions that would work perfectly with either method.

jdhoffa:
> I'm not opposed to the changes here in general, but they don't address either of the problems mentioned in #64, so I would not close that ticket with this PR

As mentioned in the original issue, I believe the two topics flagged are not actually problems. When running the workflow using the `docker` method, all of those variables are required. I would close that issue now since the README doesn't necessitate changes currently. 

However, if you want we can open an issue requesting simultaneous support for "Docker" and "RStudio" workflows of data prep.

PR #68, opened 2023-11-29
add an ISS data year parameter and set to 2021 for 22Q4

cjyetman:
(leave as 2019 for 21Q4)

can't connect to FactSet database at the moment, so I can't actually test this, but it seems like the appropriate logic


Antoine-Lalechere (approved):
LGTM

jacobvjk (approved):
can't judge the decision behind 2021 to 2019 ISS, but overall the implementation lgtm

PR #69, opened 2023-11-29
update FactSet database connect info

cjyetman:
I don't really know how/when/why these connect details changed, so would be good if @AlexAxthelm and/or @Antoine-Lalechere can validate these changes.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -4,8 +4,8 @@ default:
   masterdata_ownership_filename: ""
   masterdata_debt_filename: ""
   ar_company_id__factset_entity_id_filename: ""
-  dbname: "delta"
-  host: "data-eval-db.postgres.database.azure.com"
+  dbname: "FDS"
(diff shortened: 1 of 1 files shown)


Antoine-Lalechere (commented):
it looks good to me, but I would defer to @AlexAxthelm for that one 

mostly due to this ticket: https://dev.azure.com/RMI-PACTA/2DegreesInvesting/_workitems/edit/8919?src=WorkItemMention&src-action=artifact_link

AlexAxthelm (approved):

AlexAxthelm:
Note this is intended to be a short-lived database, and will probably not be available for next round of data prep. I am preparing a PR that will work for a long-term solution.

cjyetman:
closing then... makes sense to wait for the proper details

AlexAxthelm:
Deleting branch, since all of the relevent work has migrated to `workflow.factset`

PR #73, opened 2024-01-15
remove FactSet database code

### LINKED ISSUE #72: remove FactSet database code
If the FactSet database is dead for good (@AlexAxthelm?), we should remove all the code and parameters related to that and effectively set `update_factset: FALSE` all the time
https://github.com/RMI-PACTA/workflow.data.preparation/blob/3908094ba85dbae289e5c888612359fd23f5bedc/config.yml#L9

This would likely lead to significant simplification of the process. We should probably also deprecate and remove the relevant functions in pacta.data.preparation.

cjyetman:
closes #72 

FactSet database is gone, so this simplifies things a bit.


jdhoffa:
I have so many questions lol. Hopefully they have simple answers:

This seems absurdly dangerous to me. The 'get_*' functions are not trivial, a lot of information is encoded in them. 

Is this functionality replicated somewhere? 
How can we ensure consistency between whatever new process pulls these tables and what is pulled here/ has been pulled in the past?
Where is the new functionality stored and documented?

In any case, this PR at least requires:
- an update to the README that it is expected that these datasets are required input
- specific documentation for exactly how to reproduce each dataset, or pointer to a repo containing the updated pulling functionality 

When/ why did we stop using the FactSet database?

Otherwise LGTM 😭

jdhoffa:
Apologies if I'm freaking out for no reason. Maybe this is already documented haha

cjyetman:
As I understand it, the database is gone and it ain't coming back. The FactSet `get_*` functions live in pacta.data.preparation, so they are not removed in this PR.

Not exactly sure, but I guess the process of getting FactSet data is now encoded in a new repo that @AlexAxthelm created.

> an update to the README that it is expected that these datasets are required as input

Fair enough. I updated the README to remove the relevant FactSet settings, but I can add a note about the prep-prepared FactSet data being available in the inputs directory once I figure out exactly what that looks like.

> specific documentation for exactly how to reproduce each dataset, or pointer to a repo containing the updated pulling functionality

not sure replication instructions are feasible to detail in the README, but yeah we can link to the new FactSet dow loader repo

jdhoffa:
Yeah fair enough I think I overreacted a bit, but mainly I want to be certain that we have a smooth process for pulling/ preparing these datasets and stitching them in as inputs, since that's what they have to be now. 

So yeah then I guess minimal changes required are:
- explain FS datasets that are necessary as input in README
- link to whatever repo now pulls them, @AlexAxthelm ?

AlexAxthelm (approved):

jdhoffa (changes requested):
See comment thread

jdhoffa:
> Yeah fair enough I think I overreacted a bit, but mainly I want to be certain that we have a smooth process for pulling/ preparing these datasets and stitching them in as inputs, since that's what they have to be now.
> 
> So yeah then I guess minimal changes required are:
> 
> * explain FS datasets that are necessary as input in README
> * link to whatever repo now pulls them, @AlexAxthelm ?

It seems that the repo that now pulls this is here: https://github.com/RMI-PACTA/workflow.factset
With the `README.md` above pointing to that, we should be good to go

AlexAxthelm:
> explain FS datasets that are necessary as input in README
@jdhoffa I agree that we should document the input format of the files that are exported by `workflow.factset`.

README.md
```diff
@@ -16,10 +15,24 @@ R_CONFIG_ACTIVE=YYYYQQ
...
 Run `docker-compose up` from the root directory, and docker will build the image (if necessary), and then run the data.prep process given the specified options in the .env file.
 
 Use `docker-compose build --no-cache` to force a rebuild of the Docker image.
+
+## Required Input Files
+
+### FactSet Data
+
+Files exported by [`{workflow.factset}`](https://github.com/RMI-PACTA/workflow.factset) provide financial data to tie to production data.
+See the [`workflow.factset` README](https://github.com/RMI-PACTA/workflow.factset#exported-files) for more information on expected file format.
```
AlexAxthelm, line 29: @jdhoffa This link should work properly once https://github.com/RMI-PACTA/workflow.factset/pull/11 is approved and merged.

AlexAxthelm (dismissed):

cjyetman:
@AlexAxthelm I don't have the bandwidth today (literally and figuratively) to fix it, but...

I think these paths will need to be changed for this to work (assuming the pre-prepared FactSet data files would be placed in the same input folder as the AI files)
https://github.com/RMI-PACTA/workflow.data.preparation/blob/7c9feed63fd9230f3caabba6ad95884310b718c7/run_pacta_data_preparation.R#L72-L82

jdhoffa:
> @AlexAxthelm I don't have the bandwidth today (literally and figuratively) to fix it, but...
> 
> I think these paths will need to be changed for this to work (assuming the pre-prepared FactSet data files would be placed in the same input folder as the AI files)
> 
> https://github.com/RMI-PACTA/workflow.data.preparation/blob/7c9feed63fd9230f3caabba6ad95884310b718c7/run_pacta_data_preparation.R#L72-L82

I think the file paths would be the same? The process just drops everything into the inputs folder, so it would stay as the inputs folder. 

However, I guess we could move the lines of code logically from "pre-flight" to "input"

cjyetman:
ah crud, you're right... `data_prep_inputs_path` seems appropriate... I'm confusing myself in circles.... I was remembering some former state where I used an "intermediary" directory or something like that for the "pre-flight" files

carry on 😉

jdhoffa:
> ah crud, you're right... `data_prep_inputs_path` seems appropriate... I'm confusing myself in circles.... I was remembering some former state where I used an "intermediary" directory or something like that for the "pre-flight" files
> 
> carry on 😉

no problem. We tend to change things: fast, frequently, and profoundly
so i'd say it's pretty fair that you forget the current set-up hahahaha

still, i do think a minor improval would be to put the lines under the "input" heading, since there no longer really 'pre-flight'

AlexAxthelm (approved):

README.md
jdhoffa, line 29: Some input files are already defined explicitly in the `README.md` above (under `HOST_INPUTS_PATH`).
I would say combine these with the list above or combine the list above down here, but it's confusing to mention input datasets in two different places.

My preference would be the latter.

jdhoffa (changes requested):
1 change requested in documentation, otherwise LGTM

AlexAxthelm (approved):

AlexAxthelm:
@cjyetman can we move this as not draft?

jdhoffa:
I think both seem prudent to try to get a quick solve in, while we're here, if possible? Tracking data provenance and tracking the timestamp also seem related

cjyetman:
> I think both seem prudent to try to get a quick solve in, while we're here, if possible? Tracking data provenance and tracking the timestamp also seem related

very much related... `factset_data_timestamp` used to define what FactSet data was collected, now it's orphaned and not really related to anything specific. It would be very easy to swap in FactSet files from a different/wrong timestamp and be none the wiser.

jdhoffa:
Yeah I would tend to say this should be considered/ solved before merging this. @AlexAxthelm is there a strategy/ naming convention for the data output by [workflow.factset](https://github.com/RMI-PACTA/workflow.factset) that can help ensure the desired timestamp (and other factors, data of preparation etc.) is used here?

cjyetman:
Also now noticing some changes to what's included in the manifest that should be done here to align with the change of not downloading FactSet data in this process.

cjyetman:
Only thing left now is point 2 in [above comment](https://github.com/RMI-PACTA/workflow.data.preparation/pull/73#issuecomment-1903913904), but I think that is not achievable until/if workflow.factset is modified to use some kind of timestamped filenames.

AlexAxthelm:
> 2\. The AI datasets have their filenames encoded in the parameters, I guess primarily because their filename has their provenance baked into it. Not sure if such a thing exists with the FactSet files, but maybe it should? So it's more explicit which (version of) files we're using as input? That was part of the original design here... that the config file would live here in git tracking and be a very specific record of what was used to create any version of PACTA data... otherwise, it's pretty hard to tell _which_ FactSet files had been used and be able to replicate.

`workflow.factset` includes both the date of interest (`factset_data_timestamp`) and the date the data was pulled as part of the export path. Together these define a unique set of factset data. Does this address the concern?

Here's a config from the `feature/azure-container` branch which shows how that's used (once this PR is merged, we can go through and scrub out all the `update_factset`s).

https://github.com/RMI-PACTA/workflow.data.preparation/blob/acc9c0fa0fd66fe3f536c2d07ba37b381125b77c/config.yml#L102-L109

cjyetman:
@AlexAxthelm `factset_dataset` could probably work, though it looks like it's just in the config file but not used anywhere in the script, so hypothetically it could be set to anything and nothing would change, which suggests to me that it's not very deterministic of the output. Ideally it should be captured in the manifest file also.

cjyetman:
made the infrastructure for parameterizing the FactSet input filenames... for now they're set to the standard filenames with no timestamp for each quarterly config

AlexAxthelm:
> `factset_dataset` could probably work, though it looks like it's just in the config file but not used anywhere in the script

you're correct that it's not referenced in the `run_data_preparation.R` script, but it is in the `copy_raw_data.R`: https://github.com/RMI-PACTA/workflow.data.preparation/blob/acc9c0fa0fd66fe3f536c2d07ba37b381125b77c/ACI/copy_raw_data.R#L56-L71

Not to get too much into an explanation of #75 , but since I was trying to not change the logic of the core script in that PR, I added a script to copy the raw datafiles into a data prep inputs directory (from their home locations on AFS), and then run `run_data_preparation.R` as-is (+ some more logging). So that config param determines which FS file get copied into data prep inputs.

cjyetman:
> > `factset_dataset` could probably work, though it looks like it's just in the config file but not used anywhere in the script
> 
> you're correct that it's not referenced in the `run_data_preparation.R` script, but it is in the `copy_raw_data.R`:
> 
> https://github.com/RMI-PACTA/workflow.data.preparation/blob/acc9c0fa0fd66fe3f536c2d07ba37b381125b77c/ACI/copy_raw_data.R#L56-L71
> 
> Not to get too much into an explanation of #75 , but since I was trying to not change the logic of the core script in that PR, I added a script to copy the raw datafiles into a data prep inputs directory (from their home locations on AFS), and then run `run_data_preparation.R` as-is (+ some more logging). So that config param determines which FS file get copied into data prep inputs.

I'm pretty concerned about changes like that. As the person that originally created this repo, I've implicitly been the go-to person for figuring out what happened when something goes wrong, and the more we deviate from the path that I laid out, the less I feel comfortable with that responsibility, so you're going to have to start taking over responsibility of that if we keep moving away from a process that I'm comfortable with towards one that you're comfortable with.

jdhoffa:
Seems like we're so close here

> > > `factset_dataset` could probably work, though it looks like it's just in the config file but not used anywhere in the script
> > 
> > 
> > you're correct that it's not referenced in the `run_data_preparation.R` script, but it is in the `copy_raw_data.R`:
> > https://github.com/RMI-PACTA/workflow.data.preparation/blob/acc9c0fa0fd66fe3f536c2d07ba37b381125b77c/ACI/copy_raw_data.R#L56-L71
> > 
> > Not to get too much into an explanation of #75 , but since I was trying to not change the logic of the core script in that PR, I added a script to copy the raw datafiles into a data prep inputs directory (from their home locations on AFS), and then run `run_data_preparation.R` as-is (+ some more logging). So that config param determines which FS file get copied into data prep inputs.
> 
> I'm pretty concerned about changes like that. As the person that originally created this repo, I've implicitly been the go-to person for figuring out what happened when something goes wrong, and the more we deviate from the path that I laid out, the less I feel comfortable with that responsibility, so you're going to have to start taking over responsibility of that if we keep moving away from a process that I'm comfortable with towards one that you're comfortable with.

Sorry to be nit-picky, but we are talking about code changes in a different PR now. It is totally valid to be critical of that change, but can we keep the conversation focused on this PR for now to avoid confusion? (for my sake haha)

cjyetman:
> Sorry to be nit-picky, but we are talking about code changes in a different PR now. It is totally valid to be critical of that change, but can we keep the conversation focused on this PR for now to avoid confusion? (for my sake haha)

The relevance to this PR is that it is still uncertain how to specify which set of FactSet data one is using in a way that is traceable and repeatable.

cjyetman:
> And there is a proposed way to handle this in an unmerged PR? If so, I think maybe we move the whole conversation to that PR (and flag it as a dependency for this one)

There is a proposed way in a different PR that I don't really agree with, so seems a bit strange to let this PR be dependent on something else that may or may not ultimately exist as-is, no?

jdhoffa:
Well, that's up to you I think. 

If you think that 
> 2. The AI datasets have their filenames encoded in the parameters...
is something that needs to be addressed before this PR is merged, and if there is no suitable way to do it in the current set-up, then this PR automatically depends on something. 

If you don't want to address it before merging, then we can avoid that dependency, but we may be merging something half-baked

jdhoffa:
But, in the case that we are discussing "whatever the PR is that tries to solve data provenance". I think discussions about that PR should occur on that PR?

cjyetman:
I think that this PR is not fully baked, i.e. ready to merge, if it does not have a way of tracking the provenance of the FactSet data, which would be a regression from main.

jdhoffa:
Yup agreed. 

Then this PR on either #75 or #78 (or some as-of-yet not opened PR), and discussions about how that is solved should happen there.

AlexAxthelm:
I've added a [comment](https://github.com/RMI-PACTA/workflow.data.preparation/pull/75#issuecomment-1906054837) in #75 outlining a change to that PR which might address this discussion, but basically the idea is:

* Change from specifying a single "inputs" directory with all of the files in it, to specifying the path to each type of data (AI, FS)
* if we want to make a copy of the data inputs, use a `copy_inputs` config option.

That would let us keep the current behavior by setting the AI and FS paths both to `/inputs`, but point to the actual raw files on Azure Files when we're running on ACI (without copying to an intermediate directory in between).

cjyetman:
Implemented `asset_impact_data_path` and `factset_data_path` as @AlexAxthelm [suggested here](https://github.com/RMI-PACTA/workflow.data.preparation/pull/75#issuecomment-1906054837)

So for the way I use this repo, the expectation is that one would download the appropriate FactSet archive from Azure and unzip the tar.gz and set the `factset_data_path` to the name of the unzipped directory, e.g. "factset-pacta_timestamp-20221231T000000Z_pulled-20231221T195325Z", from which the provenance of the FactSet files can be inferred and recorded in the manifest file. Hence:

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1f25f763d9a74ed513b1bcd0e05a374f1fa9a701/config.yml#L87-L88

and

https://github.com/RMI-PACTA/workflow.data.preparation/blob/1f25f763d9a74ed513b1bcd0e05a374f1fa9a701/run_pacta_data_preparation.R#L785-L789

config.yml
```diff
@@ -1,16 +1,19 @@
 default:
-  data_prep_inputs_path: "/inputs"
   data_prep_outputs_path: "/outputs"
+  asset_impact_data_path: "/inputs"
+  factset_data_path: ""
```
AlexAxthelm, line 4:
```suggestion
  factset_data_path: "/inputs"
```

if this is `""`, then the script would expect to find those files at the current working directory (which is unlikely). defaulting to `/inputs` mirrors the current behavior of "throw everything into a single dir, and mount it to `/inputs`."

AlexAxthelm (approved):
One (non-critical) suggestion. overall LGTM.

config.yml
cjyetman: so the strategy with the config file is:
- list every possible parameter in the default config, but if it's a parameter that does not have a logical default value, e.g. one that necessarily is specific to a given quarter, then leave the default value null/blank/whatever. The intention there is that if one has not overrided the default value in their selected config, then it should fail, because there's no way this should work properly unless that has been explicitly set
- in each of the quarter specific configs, strive to minimize the amount of parameters set by inheriting as much as possible from the default config, but for parameters that are necessarily specific to that quarter/config, explicitly set them here in GH so that if one day someone comes back to question "which version of FactSet is supposed to be used to build 2022Q4 data?" it will be clearly recorded here in the config. (Though, if one manually edited that before running data.prep, the setting that they used would be captured in the manifest file... so if one was working backwards in the other direction, e.g. "what version of the FactSet data was used to build the PACTA data in this folder?", they could determine that by looking at the manifest in that folder)

With the last few commits I made here, the `factset_data_path` is required for any given quarter. That's the only way I see at the moment to have it clearly known precisely which version of the FactSet data was/should be used.

cjyetman: For further explanation.... I'd prefer that the filenames of the FactSet data files have a/the timestamp integrated into them, but in lieu of that the directory name from which they come is the only place where that is currently defined.

AlexAxthelm: That makes sense. I've opened https://github.com/RMI-PACTA/workflow.factset/issues/16, which can make good use of the changes in this PR.

cjyetman:
@AlexAxthelm @jdhoffa so I think this is ready to go now, with the caveat that once https://github.com/RMI-PACTA/workflow.factset/issues/16 lands I'd prefer to adapt to start using the FactSet data filenames as the source of truth for the FactSet timestamp

AlexAxthelm:
Approved on my side.

jdhoffa:
Approved! sorry missed this somehow

PR #74, opened 2024-01-18
Automation: Azure Container Instance

AlexAxthelm:
Contains infrastrucure to Build a docker image suitable to run on Azure Container Instances, as well as a deploy template.


AlexAxthelm:
Coincidentally, I believe that this should also address issue #3, since all the secrets would be passed in via the parameters file.

AlexAxthelm:
Closing since the PR is in the wrong direction.

PR #76, opened 2024-01-19
Change logging package, and clean logging strings

AlexAxthelm:
This PR:
* changes out the logging package from `{rlog}` to `{logger}`, 
* standardizes the format for string interpolation (using `{glue}` via `logger::formatter_glue()`)
* Cleans up the logging strings:
  * End strings with a single `.`
  * Quoting filenames

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -1,10 +1,12 @@
+logger::log_threshold(Sys.getenv("LOG_LEVEL", "INFO"))
+logger::log_formatter(logger::formatter_glue)
+
 # necessary packages -----------------------------------------------------------
 
 suppressPackageStartupMessages({
   library(pacta.data.preparation)
   library(pacta.data.scraping)
   library(pacta.scenario.preparation)
-
   library(DBI)
   library(dplyr)
   library(readr)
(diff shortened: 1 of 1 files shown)


run_pacta_data_preparation.R
```diff
@@ -1,21 +1,19 @@
+logger::log_threshold(Sys.getenv("LOG_LEVEL", "INFO"))
+logger::log_formatter(logger::formatter_glue)
```
jdhoffa, line 2: NIT: i rarely add anything to a script prior to loading package. Is it important to have this prior to loading packages in this case? (Genuinely curious)

jdhoffa (approved):
NIT comment/ question, otherwise LGTM

run_pacta_data_preparation.R
AlexAxthelm: I don't think this is hugely important, but that's probably because I'm of a mind to not load packages  and replace all the `library()` calls with explicit namespacing (ex: `dplyr::bind_rows()`).

Also, tend to think of these lines as temporary until I can find some time to close up https://github.com/daroczig/logger/issues/73

PR #78, opened 2024-01-21
add infrastructure to run on VM instead of ACI

AlexAxthelm:
This is a workaround fix, to deal with the fact that ACI only allows 16Gb of RAM on container groups with no GPUs (and we're waiting on Microsoft to approve our quota increase)

This PR adds infrastructure to spin up a VM and use it similar to how we would use ACI.

### DIFF
--- a/ACI/README.md
+++ b/ACI/README.md
@@ -25,8 +25,13 @@ Due to data prep requiring >16 GB of memory (which is the current limit for stan
 
 ```sh
 
-RESOURCEGROUP="pacta-experiments" \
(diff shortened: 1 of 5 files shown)


AlexAxthelm:
Closing this PR, since most of the unique work has been integrated as part of #142

AlexAxthelm:
For future reference, `mount_afs.sh` was included in-repo in #142, and everything else was scripts for automating invocation, which was un-automated, and listed as manual steps in README. cc @cjyetman , @jdhoffa

PR #79, opened 2024-01-22
export SQLite files by default

cjyetman:
I believe that we agreed to turn this on by default in a Tech Review call some weeks ago?


jdhoffa (approved):
¯\_(ツ)_/¯ lgtm

PR #81, opened 2024-01-29
remove `iss_emissions_year` parameter

cjyetman:
since FactSet data extraction has moved to https://github.com/RMI-PACTA/workflow.factset this is no longer used/needed


jdhoffa (approved):
lgtm

AlexAxthelm (approved):

PR #92, opened 2024-02-08
add instructions to README for running locally in RStudio

### LINKED ISSUE #67: add relevant instructions to README for running locally
The current README is relevant if one is using Docker and the [docker-compose.yml](https://github.com/RMI-PACTA/workflow.data.preparation/blob/main/docker-compose.yml) to set mounted directories correctly inside the container, but if one is not using Docker to run the script the instructions in the current README are not relevant.

I propose that we create a new section in the README with relevant instructions for running locally and clearly identify each section e.g. "Docker instructions" and "Local instructions".

cjyetman:
closes #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 ### R package dependencies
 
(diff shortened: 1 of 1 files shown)


AlexAxthelm (approved):
🚀

PR #93, opened 2024-02-08
add build preferences to .Rproj file (because it insists)

cjyetman:
Because a DESCRIPTION file has been added to this repo, RStudio seems to insist on adding these build preferences to the .Rproj file. As far as I know, they are meaningless and won't cause any problems in this context, so I'm allowing them rather than constantly fighting them to keep my git history clean.


jdhoffa (approved):
Not going to open this can of worms again, but of the reason that I usually ignore `.Rproj` XD

cjyetman:
I prefer them in the repos I maintain because I develop almost exclusively in RStudio and they define/enforce a number of worthwhile editor preferences.

PR #97, opened 2024-02-09
determine and install dependencies with {pak}

### LINKED ISSUE #96: package install in Dockerfile fails because it cannot find package "R"
https://github.com/RMI-PACTA/workflow.data.preparation/blob/c08b8fb5a139716997ae2c0e83a01757d51a9a2d/Dockerfile#L22

This seems to be the same problem discovered [here](https://github.com/RMI-PACTA/workflow.transition.monitor/pull/231#issuecomment-1921786348) and elsewhere. {renv} current dev version has been fixed to not include "R" as a R package dependency, but I suspect a better solution would be to follow the same solution as in https://github.com/RMI-PACTA/workflow.transition.monitor/pull/231, which was to drop using {renv} at all and instead use  `pak::local_deps()` or `pak::local_install_deps()`.

cjyetman:
closes #96

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update \
     libpng-dev libtiff-dev pandoc git libgit2-dev \
     && rm -rf /var/lib/apt/lists/*
 
-RUN Rscript -e 'install.packages(c("pak", "renv"))'
+RUN Rscript -e 'install.packages("pak")'
 
 COPY . /workflow.data.preparation
 
@@ -18,9 +18,7 @@ WORKDIR /workflow.data.preparation
 
 RUN Rscript -e '\
   readRenviron(".env"); \
-  non_cran_pkg_deps <- c("RMI-PACTA/pacta.scenario.preparation", "RMI-PACTA/pacta.data.preparation", "RMI-PACTA/pacta.data.scraping"); \
-  cran_pkg_deps <- setdiff(renv::dependencies()$Package, basename(non_cran_pkg_deps)); \
-  pak::pkg_install(pkg = c(non_cran_pkg_deps, cran_pkg_deps)); \
+  pak::local_install_deps(); \
   '
 
 CMD Rscript run_pacta_data_preparation.R


cjyetman:
This seems to work on my machine. Can either @jdhoffa and/or @AlexAxthelm verify that one their machines?

jdhoffa:
Same as #98 , I am not sure where I can find the relevant data prep input data to be able to test this locally.

cjyetman:
> Same as #98 , I am not sure where I can find the relevant data prep input data to be able to test this locally.

Maybe @AlexAxthelm is better suited to answer this because now all of the input files are stored on Azure, and there's two separate "rawdata" directories, and honestly I'm a bit confused myself what the structure is supposed to be and also find it rather difficult to navigate Azure to get to them. I used input files that I already had on my laptop.

jdhoffa:
> > Same as #98 , I am not sure where I can find the relevant data prep input data to be able to test this locally.
> 
> Maybe @AlexAxthelm is better suited to answer this because now all of the input files are stored on Azure, and there's two separate "rawdata" directories, and honestly I'm a bit confused myself what the structure is supposed to be and also find it rather difficult to navigate Azure to get to them. I used input files that I already had on my laptop.

Sounds good! Once that is resolved, I will review this :-) 
(not trying to block things, just trying to be more prudent in general on our PROD repos)

Dockerfile
```diff
@@ -10,17 +10,15 @@ RUN apt-get update \
     libpng-dev libtiff-dev pandoc git libgit2-dev \
     && rm -rf /var/lib/apt/lists/*
 
-RUN Rscript -e 'install.packages(c("pak", "renv"))'
+RUN Rscript -e 'install.packages("pak")'
 
 COPY . /workflow.data.preparation
```
jdhoffa, line 15: NIT: I recommend copying in `DESCRIPTION` first, installing the `deps` there, and then copying in the rest of the repository. 

That takes advantage of `Dockerfile` caching and ensuring faster builds.

jdhoffa (approved):
Built Docker image and everything installs as expected.
LGTM

Dockerfile
cjyetman: opened #109

PR #98, opened 2024-02-09
remove chrome install no longer needed for pacta.data.scraping

### LINKED ISSUE #91: remove no longer necessary install of Chrome and its dependencies
https://github.com/RMI-PACTA/workflow.data.preparation/blob/7a9a3b5b65c235e1a880ffcdeb744ec6c860677a/Dockerfile#L3-L11

@jdhoffa I believe this was added in #43 solely to support using `pacta.data.scraping::get_index_regions()`, which used to depend on {chromate} which depends on having an available Chrome or Chromium installation. Since https://github.com/RMI-PACTA/pacta.data.scraping/pull/20 changed the dependency from {chromate} to {V8}, I believe this could be removed.

cjyetman:
closes #91

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,5 @@
 FROM --platform=linux/amd64 rocker/tidyverse
 
-# install system dependencies for R packages
-RUN apt-get update \
-    && [ $(which google-chrome) ] || apt-get install -y gnupg curl \
-    && [ $(which google-chrome) ] || curl -fsSL -o /tmp/google-chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb \
-    && [ $(which google-chrome) ] || DEBIAN_FRONTEND='noninteractive' apt-get install -y /tmp/google-chrome.deb \
-    && apt-get install -y libcurl4-openssl-dev libssl-dev make libicu-dev libxml2-dev \
-    zlib1g-dev libfontconfig1-dev libfreetype6-dev libfribidi-dev libharfbuzz-dev libjpeg-dev \
-    libpng-dev libtiff-dev pandoc git libgit2-dev \
-    && rm -rf /var/lib/apt/lists/*
-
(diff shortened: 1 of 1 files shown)


jdhoffa:
Can you point me to input files that I can test this with?

cjyetman:
> Can you point me to input files that I can test this with?

It should work with any valid input files. The input files are largely irrelevant to this change because this change should only affect the ability to run `pacta.data.scraping::get_index_regions()`.

jdhoffa:
I have no input datasets on my computer ("valid" or otherwise), and am unsure where such data lives now, so I am unable to review this in its entirety. 

If you think that, for this particular PR, stepping through the `run_pacta_data_preparation.R` until the point of `pacta.data.scraping::get_index_regions()` is sufficient, then I can do that. 

But in general, I would prefer not merging PRs to PROD repos without sufficiently testing their stability (in the absence of any kind of real automated CI/CD)

cjyetman:
> If you think that, for this particular PR, stepping through the `run_pacta_data_preparation.R` until the point of `pacta.data.scraping::get_index_regions()` is sufficient, then I can do that.

Unfortunately not, because this PR does not affect the running-in-RStudio method of running data.prep at all, it only impacts whether running in the specified Docker image/container will have the necessary system dependencies.

> But in general, I would prefer not merging PRs to PROD repos without sufficiently testing their stability (in the absence of any kind of real automated CI/CD)

💯 that's why I'm hoping one or both of you can verify it too

cjyetman:
Though, if one of you has a way to step through the process in the Docker container to only test that `pacta.data.scraping::get_index_regions()` still works and you're confident/satisfied with that, I guess that would be enough.

jdhoffa:
Docker run triggered, I will let you know what happens at the scraping step

jdhoffa:
Success:
Screenshot 2024-02-09 at 12 41 28

jdhoffa (approved):

PR #107, opened 2024-02-09
Generalize input pathname in `README.md`

jdhoffa:
Used to point to `AR2023Q4`, but since the inputs are now generalized to also include factset (and likely soon to include scenarios too), makes sense to just generalize it.


cjyetman:
just to be clear, this is just in the example

cjyetman (approved):

PR #115, opened 2024-02-09
Create test-Docker-build-works.yml

cjyetman:
adds a very quick and simple test to ensure that a successful build of the Docker image is not affected by a PR

### DIFF
--- a/.github/workflows/test-Docker-build-works.yml
+++ b/.github/workflows/test-Docker-build-works.yml
@@ -7,8 +7,6 @@ jobs:
   docker-compose-build:
     runs-on: ubuntu-latest
     timeout-minutes: 25
-    env:
-      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
       - name: Checkout workflow.data.preparation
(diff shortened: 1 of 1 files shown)


AlexAxthelm:
@cjyetman We're getting close to including docker actions in https://github.com/RMI-PACTA/actions. Maybe let's just use that? (some time next week)

AlexAxthelm:
@cjyetman what are you trying to achieve with this PR? Under the current system, the docker build process requires a `.env` file with a GH PAT that has read access to some private repos, which isn't being added here. 

Possible options:
* Use docker [build secrets](https://docs.docker.com/build/building/secrets/)
* Add a step to write a `.env` file that has the `REPO_PAT` GH Actions secret
* wait until there aren't dependencies on private repos.

cjyetman:
I wanted a quick auto-test that the Docker build does not get broken by changes... because otherwise I'm testing locally once in RStudio and then once in Docker every time I change something.

cjyetman:
Yeah, I stalled on enabling the GitHub PAT and figured it would be easier to just wait till we're not using private repos anymore, e.g. the requisite repos become public.

jdhoffa (approved):
lgtm!

Issue #94, opened 2024-02-08
Consider refactoring components with strict methodological significance to `pacta.data.preparation` or `pacta.scenario.preparation`


jdhoffa:
**PROBABLY DON'T DO THIS PRIOR TO SUCCESSFUL DELIVERY OF PACTA COP CH 2024** 

Towards a world where `workflow`s handle file I/O and configuration and DevOps mainly
and `pacta.*` handles as much methodology as possible. Relates to https://github.com/RMI-PACTA/practices/issues/2

- [x] maybe refactor to `pacta.data.preparation` (done in https://github.com/RMI-PACTA/pacta.data.preparation/pull/357 and #208)
https://github.com/RMI-PACTA/workflow.data.preparation/blob/5801f959827584014019d560cb17470a4352e0dd/run_pacta_data_preparation.R#L94-L98

- [x] refactor to `pacta.scenario.preparation` ~(done in https://github.com/RMI-PACTA/pacta.data.preparation/pull/358 and #209)~ not going to do this until after a switch to pacta.scenario.data.preparation is made
https://github.com/RMI-PACTA/workflow.data.preparation/blob/5801f959827584014019d560cb17470a4352e0dd/run_pacta_data_preparation.R#L131-L146

- [x] refactor to `pacta.data.preparation` (done in https://github.com/RMI-PACTA/pacta.data.preparation/pull/359 and #210)
https://github.com/RMI-PACTA/workflow.data.preparation/blob/5801f959827584014019d560cb17470a4352e0dd/run_pacta_data_preparation.R#L175-L186

- [x] refactor to `pacta.scenario.preparation` ~(done in https://github.com/RMI-PACTA/pacta.data.preparation/pull/358 and #209)~ not going to do this until after a switch to pacta.scenario.data.preparation is made
https://github.com/RMI-PACTA/workflow.data.preparation/blob/5801f959827584014019d560cb17470a4352e0dd/run_pacta_data_preparation.R#L196-L213

- [x] refactor to `pacta.data.preparation` (done in https://github.com/RMI-PACTA/pacta.data.preparation/pull/348)
https://github.com/RMI-PACTA/workflow.data.preparation/blob/5801f959827584014019d560cb17470a4352e0dd/run_pacta_data_preparation.R#L260-L270

- [x] refactor to `pacta.data.preparation` (done in https://github.com/RMI-PACTA/pacta.data.preparation/pull/360 and #211)
[...]

PR #116, opened 2024-02-09
ensure the AI Company ID to Entity ID data has only distinct rows

### LINKED ISSUE #101: deal with `left_join()` many-to-many relationship warnings
A few `left_join()` commands in the abcd_flags section of [run_pacta_data_preparation.R](https://github.com/RMI-PACTA/workflow.data.preparation/blob/main/run_pacta_data_preparation.R) throw many-to-many relationship warnings.

In each case, we should verify if the many-to-many relationship is truly expected (conceptually). If so, these warnings can be silenced by explicitly specifying `relationship = "many-to-many"`. If not, we need to look into why that is happening and fix it.

I have identified the following lines that exhibit this behavior:

https://github.com/RMI-PACTA/workflow.data.preparation/blob/c08b8fb5a139716997ae2c0e83a01757d51a9a2d/run_pacta_data_preparation.R#L359

https://github.com/RMI-PACTA/workflow.data.preparation/blob/c08b8fb5a139716997ae2c0e83a01757d51a9a2d/run_pacta_data_preparation.R#L360

https://github.com/RMI-PACTA/workflow.data.preparation/blob/c08b8fb5a139716997ae2c0e83a01757d51a9a2d/run_pacta_data_preparation.R#L384

https://github.com/RMI-PACTA/workflow.data.preparation/blob/c08b8fb5a139716997ae2c0e83a01757d51a9a2d/run_pacta_data_preparation.R#L385

https://github.com/RMI-PACTA/workflow.data.preparation/blob/c08b8fb5a139716997ae2c0e83a01757d51a9a2d/run_pacta_data_preparation.R#L389-L392

cjyetman:
closes #101

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -243,7 +243,8 @@ factset_entity_id__ar_company_id <-
   select(
     factset_entity_id = "factset_id",
     ar_company_id = "company_id"
-  )
+  ) %>%
+  distinct()
 readRDS(factset_entity_info_path) %>%
   pacta.data.preparation::prepare_entity_info(factset_entity_id__ar_company_id) %>%
   saveRDS(file.path(data_prep_outputs_path, "entity_info.rds"))


run_pacta_data_preparation.R
```diff
@@ -243,7 +243,8 @@ factset_entity_id__ar_company_id <-
   select(
     factset_entity_id = "factset_id",
     ar_company_id = "company_id"
-  )
+  ) %>%
+  distinct()
```
jdhoffa, line 247: NIT: Should this be something we warn about? like "X input contains duplicate rows, removing"
Maybe not important.

jdhoffa (approved):
approved with nit comment

run_pacta_data_preparation.R
cjyetman: 🤷 maybe? I think that it should be expected that all of these inputs do not have random duplicate lines. 

I've already gone down the rabbit hole of imaging adding validation functions to pacta.data.preparation for all of these inputs and all of the configuration options 😈

jdhoffa: yeah i mean, in my perfect world, once https://github.com/RMI-PACTA/workflow.data.preparation/issues/94 is closed, I would imagine just adding a simple "warning" message in the wrapper function whenever this kind of wonkiness happens would be sufficient

PR #117, opened 2024-02-10
git ignore .Rbuildignore

cjyetman:


jdhoffa (approved):

AlexAxthelm:
After the fact, but why are we ignoring this file?

cjyetman:
> After the fact, but why are we ignoring this file?

so that it doesn't get accidentally pushed or suggest when making a commit... much like the `.env` and the `.DS_Store` files already in there

jdhoffa:
Yeah wait actually I agree with @AlexAxthelm , I don't think this file should be ignored... I reviewed this too hastily...

See: https://github.com/RMI-PACTA/r2dii.data/blob/main/.Rbuildignore

It gives critical info to someone trying to build the package. If we ignore it, then when someone tries to build the package from source, it will try to bundle in `run_data_preparation.R` etc.

cjyetman:
This repo is not an R package, it's a workflow. Because I added the `DESCRIPTION` file, RStudio starts doing some annoying automatic stuff, like adding build options to the `.Rproj` file (which I just succumbed to so it would stop bothering me in #93). It also adds a `.Rbuildignore` file whenever you open the project, so this will prevent that or at least stop it from being annoying.

But seriously, are these the type of things worth debating?

cjyetman:
> > But seriously, are these the type of things worth debating?
> 
> For this one, since it's not actually an R package, probably not. but in general, the `.Xignore` files, such as `.Rbuildignore` and `.dockerignore` can have huge impacts on build processes. (I'm annoyed that they're dotfiles).
> 
> As a quick example, `.dockerignore` excludes files from the docker build context, which means that when you have a line like this (currently on `main`):
> 
> https://github.com/RMI-PACTA/workflow.data.preparation/blob/ca2d53e2afceeca2b6219af8a9b3ba6e258ae5ac/Dockerfile#L20
> 
> if you don't have a `.dockerignore` defined, _everything_ in the build context gets pulled into the image, which means that if you happen to have a copy of the rawdata inputs and outputs in the same directory, those get added. As does the `.git` directory, meaning that ever time you commit, you're invalidating the cache (and these problems multiply each other).

I would very much support adding a `.dockerignore` file. I will open [an issue](https://github.com/RMI-PACTA/workflow.data.preparation/pull/128) and tag you in it, because honestly I don't know too much about it. The `.Rbuildignore` file I think is rather inconsequential in this repo.

PR #122, opened 2024-02-11
don't bind mount the repo directory that gets copied in

cjyetman:
depends on #99 

not sure why this was being done since the repo gets permanently copied into the Docker image. @jdhoffa?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,3 @@ services:
       - type: bind
         source: ${HOST_OUTPUTS_PATH}
         target: /outputs
-      - type: bind
-        source: .
-        target: /workflow.data.preparation


jdhoffa:
I don't remember... 
but let me actually try running this before you merge to see if it breaks anything.

jdhoffa:
Unable to test this with the input data that @cjyetman provided until #99 is merged.
Given that there are currently 10 open PRs on this PROD repository, I am going to slow down reviewing. 

I will review this once some outcome has come from #99

cjyetman:
good point... I will mark this as dependent on those PRs and ping you once they've merged

AlexAxthelm:
> not sure why this was being done since the repo gets permanently copied into the Docker image.

This is a pattern I've used before during development phases, especially for images with long build times. Mounting in the repo allows for (minor) code changes to be tested _in-situ_ and run, either interactively or as a script.

In any case, this is a development pattern, and probably shouldn't have be committed.

cjyetman:
@jdhoffa you wanna give this a go again now that #99 has been merged? based on @AlexAxthelm's comment and my own testing, I'm very confident this is the right thing to do and it will work if you have the proper input data and they're properly specified in the config

jdhoffa:
It is running!
Screenshot 2024-02-13 at 13 34 37

jdhoffa:
> So it runs. The process is killed after the `financial_data.rds` step because of an `OOM` exception, so I can't say anything beyond that really... : Screenshot 2024-02-13 at 13 41 46

@cjyetman unless you think that is sufficient to prove the safety of this PR?

cjyetman:
🤷🏻 I've run it with this on my machine (in Docker) at least 5 times successfully, but I did modify the config file to only use IPR2021 scenario (presumably you get OOM because you're processing all the scenarios).

I guess this makes it pretty clear why it's so difficult to maintain something that is meant to run in different contexts, but no one person can run it in all of those contexts. 🤔

cjyetman:
though.... if it runs at all @jdhoffa, that kinda proves to me this PR works, and contextually one can see from the diff that it can't/shouldn't change anything else

jdhoffa (approved):
Approving since @cjyetman affirms that running part of this workflow through Docker is enough of to constitute success here. Though I have not been able to test it to an adequate degree of satisfaction for myself. 

NIT:
* I have 16GB of RAM, I don't consider that "very resource-constrained"... but also a bit of an absurd thing to bring up, as it is just the reality of our dev team set-up. "Get a new computer" is not really an appropriate solution. 
* Crashing at `financial_data` is consistent with where it crashed for me locally 1 year ago, and why I started using Azure

And a general FYI, I will not approve future PRs with larger structural changes on this repo until one of the following things becomes true:
* I can run a complete test of the PR with real data on Azure
* I can run a complete test of the PR with real data on my local machine
* I can run a complete test of the PR with demo/ fake data on either
* (Preferred) CI/CD exists to automatically build and run the PR with demo/ fake data (relates to #6)

I am not trying to block things for no reason, but I feel that we are veering into unnecessarily dangerous territory on a crucial PROD repo, and I want to address that promptly, as I am getting further and further away from feeling comfortable when I click "approve"...

You may choose to not request me for review, but I don't want to just "rubber stamp" things blindly and without proof because I think the code changes look ok...

cjyetman:
> * I have 16GB of RAM, I don't consider that "very resource-constrained"... but also a bit of an absurd thing to bring up, as it is just the reality of our dev team set-up. "Get a new computer" is not really an appropriate solution.

What I mean is.... nothing in the process up to the point of processing the financial_data is resource intensive, and I would expect that a machine with 8GB of RAM should be able to run that much of it. "constrained" could be because of other processes running on your machine at the same time too.

jdhoffa:
> What I mean is.... nothing in the process up to the point of processing the financial_data is resource intensive, and I would expect that a machine with 8GB of RAM should be able to run that much of it. "constrained" could be because of other processes running on your machine at the same time too.

Ok. If you believe that my system constraints are the critical problem here and want to pair with me to find a solution to that (that doesn't involve buying a new computer) then I am open to that.

cjyetman:
> Ok. If you believe that my system constraints are the critical problem here and want to pair with me to find a solution to that (that doesn't involve buying a new computer) then I am open to that.

I don't believe anything, I'm quite curious because I have no idea what would be the constraint there. So happy to investigate that with you.

> The point I am making is that I cannot adequately review ANY PR opened to this repo if I can't run the API. So requesting my review is sorta pointless and more of a "feel good" check until that gets resolved.

I feel the same way. My only resolution to that is to only make PRs that I can reason about with logic alone. It's absolutely impossible/infeasible for any of to run/test every change completely, in all expected contexts, with all expected inputs, etc. It's a very difficult situation... I totally agree.

AlexAxthelm:
>  I'm quite curious because I have no idea what would be the constraint there

The biggest constraint is that these lines (when run against prod-sized data)

https://github.com/RMI-PACTA/workflow.data.preparation/blob/6152d04cb718f7a43ffe333f12ac892d1a342ab6/run_pacta_data_preparation.R#L231-L233

require more than 16GB of memory. I've opened https://github.com/RMI-PACTA/pacta.data.preparation/issues/334 to target the issue at its source.

cjyetman:
> This is precisely what having a robust CI/CD pipeline is designed to avoid. Getting that running in this repo is a priority.

I don't think anyone here as argued against that

PR #124, opened 2024-02-11
avoid unnecessary intermediate scenario files

cjyetman:
I believe we originally created these scenario CSVs to faithfully replicate the setup for the former data.prep process, but as far as I know these CSVs are not useful for anything once the current data.prep is completely. In this PR, I move that process further down and rather than export those intermediate files just process the in memory data objects.

@jdhoffa hope you can review this thoroughly and agree

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -84,8 +84,6 @@ factset_iss_emissions_data_path <-
 
 # pre-flight filepaths ---------------------------------------------------------
 
-scenarios_analysis_input_path <- file.path(asset_impact_data_path, "Scenarios_AnalysisInput.csv")
-scenario_regions_path <- file.path(asset_impact_data_path, "scenario_regions.csv")
 currencies_data_path <- file.path(asset_impact_data_path, "currencies.rds")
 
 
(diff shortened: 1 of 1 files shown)


cjyetman:
While I would be happy to see https://github.com/RMI-PACTA/workflow.scenario.preparation/issues/5 completed and leveraged here, I don't see why these are mutually exclusive. This PR doesn't change the process really, just when (later in the workflow) and where (in-memory rather than writing a temporary file) it happens. Do you see something in this PR that would prevent you from doing https://github.com/RMI-PACTA/workflow.scenario.preparation/issues/5 in the future?

jdhoffa:
Because doing things in memory would remove my entry-point into the workflow. Not a huge issue, I will just need to re-introduce reading the `.csv` again relatively soon.

cjyetman:
> Because doing things in memory would remove my entry-point into the workflow. Not a huge issue, I will just need to re-introduce reading the `.csv` again relatively soon.

In that case, I would actually strongly prefer merging this, because https://github.com/RMI-PACTA/workflow.scenario.preparation/issues/5 will require a refactoring/removal of the scenario processing code in https://github.com/RMI-PACTA/workflow.data.preparation/blob/main/run_pacta_data_preparation.R (which will be somewhat easier after this condenses it into a single block) and it will likely require adding a new file path to the config file etc.

jdhoffa (approved):
¯\_(ツ)_/¯ ok

Issue #118, opened 2024-02-10
add checks for valid config paramaters


cjyetman:
Numerous parameters are read in through the config file (see below). While the file paths/names are ultimately tested for existence, there are no basic validity checks on the parameters, e.g. checks that necessary parameters exist, that parameters are not `NULL`, that parameters are not `""`, and/or that parameter values are within a set of valid values. This could be achieved in the [run_pacta_data_preparation.R](https://github.com/RMI-PACTA/workflow.data.preparation/blob/main/run_pacta_data_preparation.R) script, or more (maybe better?) it could be achieved through a set of validation functions added in [pacta.data.preparation](https://github.com/RMI-PACTA/pacta.data.preparation).

https://github.com/RMI-PACTA/workflow.data.preparation/blob/a349890cab43dea90bc3f20cdf33a7cf6c8fc0c3/run_pacta_data_preparation.R#L20-L59

AlexAxthelm:
@cjyetman, you're basically looking for checks on the config, where it would be something along the lines of:

```r
#ugly code
if (
  isnull(config$asset_impact_data_path) || 
  config$asset_impact_data_path == "" || 
  !dir.exists(config$asset_impact_data_path) 
) {
  stop("some message")
}
```

There isn't anything inherently wrong with this approach, but it's going to get messy to maintain. 

I really think this might be time to look at a different mechanism for managing our configs. I know I've pitched JSON Schema as a cure for a lot of our problems before, but this kind of validation is exactly what it's built for.

consider this snippet, which would handle those checks (from the current `main` in workflow.pacta, with only the `jsonvalidate::` line added

```r
logger::log_trace("Determining configuration file path")
cfg_path <- commandArgs(trailingOnly = TRUE)
if (length(cfg_path) == 0 || cfg_path == "") {
  logger::log_warn("No configuration file specified, using default")
  cfg_path <- "input_dir/default_config.json"
}
logger::log_debug("Loading configuration from file: \"{cfg_path}\".")
jsonvalidate::(cfg_path, schema_path) # this is the new line
cfg <- fromJSON(cfg_path)
```

then the same code from above would be

```jsonc
{
// ... json schema preamble
  "type": "object"
  "properties": {
    "asset_impact_data_path": {
      "description": "path to raw data files from Asset Impact. See README for more information."
      "type": "string"
      "regex": "[^\0]+" // not a great regex, but at least flags illegal characters in a POSIX path.
    },
    //... another example
[...]

Issue #108, opened 2024-02-09
update explicit FactSet filenames once they are known


cjyetman:
Following #99, the expectation is that for any given timestamp config set, e.g. `2022Q4`, each FactSet file will be explicitly named and the filename will include a very precise timestamp indicating when it was made, very much like how the Asset Impact files have been done previously.

In order to know what the those timestamped names should be for a given config set, presumably a FactSet data dump would have needed to be complete and it would have had to have passed through our QA process and generally people would have had to sign off on it as "this is the FactSet data that should be used for 2022Q4".

Since we're not there yet, this is an issue to remind us to do that once it is ready.

PR #125, opened 2024-02-11
compute FactSet timestamp/s from filenames

cjyetman:
depends on #99 

This PR computes the FactSet timestamp/s from the filenames of all input FactSet data files, and adds that info to the manifest. Note that it would be possible to use FactSet data files from different timestamps... in which case all unique timestamps would be added to the manifest as an array. In the future, e.g. in #118, we could consider giving an error or warning if all of these timestamps do not match.

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -102,6 +102,16 @@ logger::log_info(
 
 scenario_raw_data_to_include <- lapply(scenario_raw_data_to_include, get, envir = asNamespace("pacta.scenario.preparation"))
 
+factset_timestamp <-
+  unique(sub("_factset_.*[.]rds$", "", c(
+    factset_financial_data_filename,
+    factset_entity_info_filename,
+    factset_entity_financing_data_filename,
+    factset_fund_data_filename,
+    factset_isin_to_fund_table_filename,
+    factset_iss_emissions_data_filename
+  )))
+
 
 # check that everything is ready to go -----------------------------------------
 
@@ -785,7 +795,7 @@ parameters <-
     ),
     timestamps = list(
       imf_quarter_timestamp = imf_quarter_timestamp,
-      factset_data_identifier = basename(factset_data_path),
+      factset_data_identifier = factset_timestamp,
       pacta_financial_timestamp = pacta_financial_timestamp
     ),
     scenarios = list(


jdhoffa:
As with #99 , I am not the right person to review this.

cjyetman:
@AlexAxthelm?

run_pacta_data_preparation.R
```diff
@@ -103,6 +103,16 @@ logger::log_info(
 
 scenario_raw_data_to_include <- lapply(scenario_raw_data_to_include, get, envir = asNamespace("pacta.scenario.preparation"))
 
+factset_timestamp <-
+  unique(sub("_factset_.*[.]rds$", "", c(
```
AlexAxthelm, line 107: Style question: I've noticed you prefer using groups to denote literal `.`, rather than escaping (`[.]` vs `\\.`). Is there a reason for this?

I don't think we're particularly worried about optimizing our regexes, but I believe there is a small performance hit for groups over escaped characters.

run_pacta_data_preparation.R
```diff
@@ -103,6 +103,16 @@ logger::log_info(
...
+    factset_financial_data_filename,
+    factset_entity_info_filename,
+    factset_entity_financing_data_filename,
+    factset_fund_data_filename,
+    factset_isin_to_fund_table_filename,
+    factset_iss_emissions_data_filename
+  )))
+
```
AlexAxthelm, line 115: What happens if the `unique` returns a length more than 1? Suggest putting a check here (or somewhere) to at least warn if more than one timestamp is found (in case user doesn't check the exported `parameters`, below)

AlexAxthelm (changes requested):
Looks good, but would like stronger check that all the filenames are as expected.

run_pacta_data_preparation.R
cjyetman: Definitely just a personal preference. I find it easier to read and harder to screw up than managing doubled backslashes to avoid R string escapes. Something I picked up from maintaining the regexes in countrycode.

run_pacta_data_preparation.R
cjyetman: See original PR description

cjyetman:
> Looks good, but would like stronger check that all the filenames are as expected.

I prefer to have PRs that are hyper-focused because I feel it's easier for one to reason about what the impact might be, and there's a separate issue (#108) and branch (in the R package) to implement a way of validating proper configuration options. But if you're adamant about having it here I can add something temporary and remove it when the config-validation thing lands.

run_pacta_data_preparation.R
cjyetman: e.g. one can copy-paste this regex into a regex validator, but one with a R-style backslash won't work the same

AlexAxthelm (approved):

AlexAxthelm:
Looks good to me. Sounds like we should move #118 up the list a bit.

Issue #100, opened 2024-02-09
save pre-flight files to a temp directory


cjyetman:
Currently, pre-flight files are saved to the inputs directory, before ultimately the properly formatted RDS versions of them are saved to the outputs directory. I think it might be safer to initially save the pre-flight files to a temp directory so that the inputs directory could potentially be read-only. @jdhoffa and @AlexAxthelm do you have any concerns with that?

jdhoffa:
I think the `pre-flights` were (if I'm not mistaken) most useful because it was nice to have a cache for the FactSet and Scenario data. 

Since it seems like we are moving towards those datasets being pulled/ stored/ versioned separately anyway (https://github.com/RMI-PACTA/workflow.factset and https://github.com/RMI-PACTA/workflow.scenario.preparation)

Since `pacta.scenario.preparation` offers a stable and versioned view of the scenario input datasets (and since FactSet is long gone), i think you're safe to do this now.

cjyetman:
Given the discussion [beginning here](https://github.com/RMI-PACTA/workflow.data.preparation/pull/127#issuecomment-1939155619), I'm closing this in favor of a strategy to eventually (optionally) copy all input files (including any "pre-flight" files) to a specific directory for archival (notably *not* a temp directory, and notably *not* a/the (possibly read-only) input directory).

PR #127, opened 2024-02-11
save currencies data only to output path

cjyetman:
To avoid writing to the inputs directory, initially save the pre-flight currencies data to an in-memory object and only write it to the outputs directory.

supersedes #100

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -88,6 +88,7 @@ scenarios_analysis_input_path <- file.path(asset_impact_data_path, "Scenarios_An
 scenario_regions_path <- file.path(asset_impact_data_path, "scenario_regions.csv")
 currencies_data_path <- file.path(data_prep_outputs_path, "currencies.rds")
 
+
 # computed options -------------------------------------------------------------
 
 relevant_years <- sort(
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -218,7 +218,7 @@ logger::log_info("Scenario data prepared.")
 
 logger::log_info("Saving file: \"currencies.rds\".")
 currencies %>%
-  saveRDS(file.path(data_prep_outputs_path, "currencies.rds"))
+  saveRDS(currencies_data_path)
 
 
 # financial data output --------------------------------------------------------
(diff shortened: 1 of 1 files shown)


AlexAxthelm (approved):
given the discussion in #100, if we end up saving a copy of "raw inputs to data prep" somewhere, would we want to include a copy of this file there?

I think for reproducibility, it's a good idea to have a copy of the files we used for a particular data prep

cjyetman:
> given the discussion in #100, if we end up saving a copy of "raw inputs to data prep" somewhere, would we want to include a copy of this file there?
> 
> I think for reproducibility, it's a good idea to have a copy of the files we used for a particular data prep

That was the original idea of it... but truth be told, the `currencies.rda` file saved in the "raw inputs" has always been the same exact thing as the `currencies.rda` file saved in the outputs. So, I questioned this too, but I came to the conclusion that the benefit of *not* writing to the inputs folder (e.g. it could potentially be read-only, which is probably preferable and might even be necessary in some context) I think outweighs the benefit of making a duplicate of the file.

cjyetman:
> > given the discussion in #100, if we end up saving a copy of "raw inputs to data prep" somewhere, would we want to include a copy of this file there?
> > I think for reproducibility, it's a good idea to have a copy of the files we used for a particular data prep
> 
> That was the original idea of it... but truth be told, the `currencies.rda` file saved in the "raw inputs" has always been the same exact thing as the `currencies.rda` file saved in the outputs. So, I questioned this too, but I came to the conclusion that the benefit of _not_ writing to the inputs folder (e.g. it could potentially be read-only, which is probably preferable and might even be necessary in some context) I think outweighs the benefit of making a duplicate of the file.

#100 was my thoughts on dealing with this a few days ago, but at the moment it just seems mostly unnecessary

cjyetman:
An alternative (I think maybe you suggested something like this somewhere recent @AlexAxthelm) could be copying all the "inputs" to a separate "archive" directory, versus the "inputs", "outputs", or a temporary directory?

AlexAxthelm:
> An alternative (I think maybe you suggested something like this somewhere recent @AlexAxthelm) could be copying all the "inputs" to a separate "archive" directory, versus the "inputs", "outputs", or a temporary directory?

The naming structure that I've been playing around with is:

- `rawdata` for input files
- `data_prep_inputs` for copies of `rawdata`, and files that are downloaded as part of data prep (currencies, indicies, etc).
- `outputs`

cjyetman:
Since you have something incoming (I think) implementing an optional `copy_data` option, can we merge this and implement the data inputs archiving with that?

cjyetman:
> Since you have something incoming (I think) implementing an optional `copy_data` option, can we merge this and implement the data inputs archiving with that?

or I could implement that separately in a mini-PR?

AlexAxthelm:
Let's merge this one in, and when we add that functionality, we'll add currencies to that then.

PR #129, opened 2024-02-13
add `.dockerignore` file

### LINKED ISSUE #128: add `.dockerignore` file
> For this one, since it's not actually an R package, probably not. but in general, the `.Xignore` files, such as `.Rbuildignore` and `.dockerignore` can have huge impacts on build processes. (I'm annoyed that they're dotfiles).
> 
> As a quick example, `.dockerignore` excludes files from the docker build context, which means that when you have a line like this (currently on `main`):
> 
> https://github.com/RMI-PACTA/workflow.data.preparation/blob/ca2d53e2afceeca2b6219af8a9b3ba6e258ae5ac/Dockerfile#L20
> 
> if you don't have a `.dockerignore` defined, _everything_ in the build context gets pulled into the image, which means that  if you happen to have a copy of the rawdata inputs and outputs in the same directory, those get added. As does the `.git` directory, meaning that ever time you commit, you're invalidating the cache (and these problems multiply each other).
> 

_Originally posted by @AlexAxthelm in https://github.com/RMI-PACTA/workflow.data.preparation/issues/117#issuecomment-1938949137_

https://docs.docker.com/build/building/context/#dockerignore-files

cjyetman:
closes #128 

based on https://github.com/RMI-PACTA/workflow.pacta/blob/main/.dockerignore with the following differences
- allowing markdown files because I think it is reasonably relevant to have the license and README markdown files baked into the image along with the associated code
- excluding the `docker-compose.yml` because that's not relevant inside the image (and I wonder if in some cases it could be a security concern, e.g. if one did not want local path specifications to be revealed inside the image)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,10 @@
-**/.git
-**/.github
-**/.gitignore
-**/Dockerfile
-**/*.Rproj
-docker-compose.yml
+# ignore everything
+*
+# include certain files
+!config.yml
+!DESCRIPTION
+!Dockerfile
+!LICENSE
+!LICENSE.md
+!README.md
+!run_pacta_data_preparation.R
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+**/.git
+**/.github
+**/.gitignore
+**/Dockerfile
+**/*.Rproj
+docker-compose.yml
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 # ignore everything
 *
 # include certain files
+!.env
 !config.yml
 !DESCRIPTION
 !Dockerfile
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,6 @@
 !.env
 !config.yml
 !DESCRIPTION
-!Dockerfile
 !LICENSE
 !LICENSE.md
 !README.md


jdhoffa (dismissed):

.dockerignore
```diff
@@ -0,0 +1,6 @@
+**/.git
+**/.github
+**/.gitignore
+**/Dockerfile
+**/*.Rproj
+docker-compose.yml
```
AlexAxthelm, line 6: Given that local testing is a common pattern with this repo, with some very large files that could get baked in, I'm thinking that a preferable pattern for this dockerignore is to exclude _everything_ and then add back in the few files that we do want to include.

```suggestion
# ignore everything
*
# include certain files
!config.yml
!DESCRIPTION
!Dockerfile
!LICENSE
!LICENSE.md
!README.md
!run_pacta_data_preparation.R
```

AlexAxthelm (dismissed):
Your implementation works, but please see my comment.

.dockerignore
cjyetman: personally, I don't target input or output directories in/under the root directory when using locally, but fair enough, totally ok with that strategy

however...
- is it necessary/desirable to (explicitly) include the Dockerfile inside the Docker image?
- I believe it is necessary to include the `.env` file (because the desired config set is specified there, not just for the mount directories and the GitHub PAT)

@AlexAxthelm you ok with those changes?

jdhoffa (approved):

.dockerignore
AlexAxthelm: > * is it necessary/desirable to (explicitly) include the Dockerfile inside the Docker image?

Hmm. I don't know. Try taking it out and see if it builds?

> * I believe it is necessary to include the `.env` file (because the desired config set is specified there, not just for the mount directories and the GitHub PAT)

Good catch.

AlexAxthelm (approved):

cjyetman:
I did test this on my machine (Docker build and compose up) a couple times, so it appears the `Dockerfile` is not needed inside the Docker image.

PR #134, opened 2024-02-15
remove IPR from default 2022Q4 config

cjyetman:
since we removed TDM, IPR scenario is no longer desired in the default 2022Q4 config (confirmed with @AlexAxthelm @NAYRA-HERRERA and Nicholas)


jdhoffa (approved):
Yay!

PR #137, opened 2024-02-15
use new `factset_manual_pacta_sector_override.rds` from `workflow.factset`

cjyetman:
- [x] depends on https://github.com/RMI-PACTA/workflow.factset/pull/47

must be coordinated with https://github.com/RMI-PACTA/pacta.data.preparation/pull/338

uses the same strategy as #133 did to add the new `factset_industry_map_bridge.rds`, including the complicated process of modifying the function signature of `prepare_entity_info()` and all that flows from that, which necessitates coordination with the connected PR in pacta.data.preparation and a dev version bump

note: as was done with the `factset_industry_map_bridge`, I load the `factset_manual_pacta_sector_override.rds` in the "intermediary files" section of the script

This tells the workflow to use the new `factset_manual_pacta_sector_override.rds` from the FactSet inputs rather than using `pacta.data.preparation::factset_manual_pacta_sector_override` which is set to be deprecated/removed in https://github.com/RMI-PACTA/pacta.data.preparation/issues/321

⚠️ in this PR, I have temporarily set the default 2022Q4 config `factset_manual_pacta_sector_override_filename` to `timestamp-20230123T000000Z_pulled-20000101T000002_factset_manual_sector_override.rds` (which was the first version of this new file that I have had access) to facilitate testing/verifying that this PR works (similar to what has been done with the other FactSet filenames), but these default filenames have not been determined/approved and will need to change once they have been identified, a task which is acknowledged and being tracked in [update explicit FactSet filenames once they are known #108](https://github.com/RMI-PACTA/workflow.data.preparation/issues/108).

For testing, this implies that one has put this `timestamp-20230123T000000Z_pulled-20000101T000002_factset_manual_sector_override.rds` file in the same inputs directory as the other FactSet input files.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,7 +33,7 @@ Imports:
     dbplyr,
     dplyr,
     logger,
-    pacta.data.preparation (>= 0.1.0.9001), 
+    pacta.data.preparation (>= 0.1.0.9002), 
     pacta.data.scraping, 
     pacta.scenario.preparation, 
     readr, 
--- a/README.md
+++ b/README.md
@@ -85,3 +85,4 @@ The required files are:
 - factset_iss_emissions.rds
 - factset_issue_code_bridge.rds
 - factset_industry_map_bridge.rds
+- factset_manual_pacta_sector_override.rds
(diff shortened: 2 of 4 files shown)


AlexAxthelm (approved):

jdhoffa (approved):
lgtm!

PR #145, opened 2024-02-18
Add current working config for 2022q4

### LINKED ISSUE #108: update explicit FactSet filenames once they are known

### LINKED ISSUE #143: Add PROD 2022Q4 configuration
Add configuration with final files used in PROD for 2022Q4 dataset

AlexAxthelm:
closes #143 
closes #108 

Leaving as Draft until data outputs based on this config have been evaluated.


AlexAxthelm:
Marking as ready for review, following @Antoine-Lalechere's comments on Teams.

config.yml
```diff
@@ -40,13 +40,13 @@ desktop:
   factset_data_path: "./factset_inputs"
 
 docker:
+  inherits: 2022Q4
   data_prep_outputs_path: "/mnt/outputs"
-  asset_impact_data_path: "/inputs"
-  factset_data_path: "/inputs"
+  asset_impact_data_path: "/mnt/inputs"
+  factset_data_path: "/mnt/inputs"
```
cjyetman: this will require a change in `docker-compose.yml`

cjyetman: never mind, somehow I keep seeing an old diff

config.yml
```diff
@@ -39,14 +39,8 @@ desktop:
...
   factset_data_path: "./factset_inputs"
 
-docker:
-  data_prep_outputs_path: "/mnt/outputs"
-  asset_impact_data_path: "/inputs"
-  factset_data_path: "/inputs"
-
 
 2021Q4:
-  inherits: docker
```
cjyetman, line 49: let's leave these in, otherwise the configs that don't have a matching `202xQ4_docker` are broken

config.yml
```diff
@@ -106,19 +99,18 @@ docker:
...
-  factset_fund_data_filename: "timestamp-20221231T000000Z_pulled-20240207T161053Z_factset_fund_data.rds"
-  factset_isin_to_fund_table_filename: "timestamp-20221231T000000Z_pulled-20240207T161053Z_factset_isin_to_fund_table.rds"
-  factset_iss_emissions_data_filename: "timestamp-20221231T000000Z_pulled-20240207T161053Z_factset_iss_emissions.rds"
-  factset_issue_code_bridge_filename: "test-from-fds-test-20240207-03-postgres_factset_issue_code_bridge.rds"
-  factset_industry_map_bridge_filename: "timestamp-20230123T000000Z_pulled-20000101T000001_factset_industry_map_bridge.rds"
-  factset_manual_pacta_sector_override_filename: "timestamp-20230123T000000Z_pulled-20000101T000002_factset_manual_sector_override.rds"
+  masterdata_ownership_filename: "2023-06-18/2023-06-18_AI_RMI Bespoke_Company Data Products_masterdata_ownership_2022Q4.csv"
+  masterdata_debt_filename: "2023-06-18/2023-06-18_AI_RMI Bespoke_Company Data Products_masterdata_debt_2022Q4.csv"
+  ar_company_id__factset_entity_id_filename: "2023-02-15/2023-02-15_AI_RMI_Bespoke_Company_Data_Products_Company_ID_List_2022Q4.csv"
```
cjyetman, line 104: I really want to avoid this, at least for now. As a compromise (as opposed to insisting that you rearrange files on Azure to comply with the input expectations here), can you leave just the filenames in the `2022Q4` config, and then also add them with the paths in the context-specialized `2022Q4_docker` or maybe `2022Q4_azure` config. That way you can set the config the way you need, but the primary `2022Q4` config stays context-agnostic.

(with a minor warning that I've never messed around putting paths in the filenames, so I can't say from experience that this is totally safe)

config.yml
```diff
@@ -106,19 +99,18 @@ docker:
...
+  factset_entity_financing_data_filename: "timestamp-20221231T000000Z_pulled-20240217T134528Z_factset_entity_financing_data.rds"
+  factset_fund_data_filename: "timestamp-20221231T000000Z_pulled-20240217T134528Z_factset_fund_data.rds"
+  factset_isin_to_fund_table_filename: "timestamp-20221231T000000Z_pulled-20240217T134528Z_factset_isin_to_fund_table.rds"
+  factset_iss_emissions_data_filename: "timestamp-20221231T000000Z_pulled-20240217T134528Z_factset_iss_emissions.rds"
+  factset_issue_code_bridge_filename: "timestamp-20221231T000000Z_pulled-20240217T134528Z_factset_issue_code_bridge.rds"
+  factset_industry_map_bridge_filename: "timestamp-20221231T000000Z_pulled-20240217T134528Z_factset_industry_map_bridge.rds"
+  factset_manual_pacta_sector_override_filename: "timestamp-20221231T000000Z_pulled-20240217T134528Z_factset_manual_sector_override.rds"
   imf_quarter_timestamp: "2022-Q4"
   pacta_financial_timestamp: "2022Q4"
   market_share_target_reference_year: 2022
```
cjyetman, line 116: This for sure needs all the other content parameters like `tech_exclude` etc. You probably started this PR before I stripped all the values from the default config. Annoying a bit I admit, but it's very much by design because there's really no reasonable, methodologically sound way for us to specify those default values across many years.

config.yml
```diff
@@ -133,3 +125,9 @@ docker:
   scenario_geographies_list: ["Global", "NonOECD", "OECD"]
   global_aggregate_scenario_sources_list: ["WEO2022"]
   global_aggregate_sector_list: ["Power"]
+
+2022Q4_docker:
+  inherits: 2022Q4
+  data_prep_outputs_path: "/mnt/outputs"
+  asset_impact_data_path: "/mnt/asset-impact"
+  factset_data_path: "/mnt/factset-extracted/factset-pacta_timestamp-20221231T000000Z_pulled-20240217T134528Z"
```
cjyetman, line 133: These require a change to `docker-compose.yml`. Maybe that's in another PR, but I'd rather not merge PRs that depend on another PR until the other PR is merged. So I guess either this PR needs to wait, or it needs to import the change here.

cjyetman: in order to keep PRs focussed, I'd prefer that this PR wait (until the necessary infrastructure is already merged)

config.yml
AlexAxthelm: Green techs and all those are here. They just aren't marked in the diff.

config.yml
AlexAxthelm: Added changes to `docker-compose.yml` to this PR.

cjyetman (approved):

PR #149, opened 2024-02-18
Feat/147 update calls to write manifest

### LINKED ISSUE #147: Update calls to `write_manifest()`
Update calls to `write_manifest()` to account for https://github.com/RMI-PACTA/pacta.data.preparation/issues/340

AlexAxthelm:
Update run_data_preparation to use changed function in https://github.com/RMI-PACTA/pacta.data.preparation/pull/341.

Also simplifies somewhat the logic around the list of input filepaths.

Should be merged after that, but coordinated closely.

**Note:** as-written, this includes #150, but reverting back to [35fd720](https://github.com/RMI-PACTA/workflow.data.preparation/pull/149/commits/35fd7200fac09024d96b93aa568b0b9316cf3af5) removes that dependency.

Also as written, depends on #153, as this does a similar simplification from `list.files()` to explicit filepath lists for the archive export feature defined in that PR. Reverting [1b8cd49](https://github.com/RMI-PACTA/workflow.data.preparation/pull/149/commits/1b8cd496ed4ffac7743dfb6030182fe5a80fff18) removes that connection.

- [x] Depends on #150 
- [x] Depends on #153 

Closes #147

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -45,6 +45,7 @@ factset_isin_to_fund_table_filename <- config$factset_isin_to_fund_table_filenam
 factset_iss_emissions_data_filename <- config$factset_iss_emissions_data_filename
 factset_issue_code_bridge_filename <- config$factset_issue_code_bridge_filename
 factset_industry_map_bridge_filename <- config$factset_industry_map_bridge_filename
+factset_manual_pacta_sector_override_filename <- config$factset_manual_pacta_sector_override_filename
 update_currencies <- config$update_currencies
 export_sqlite_files <- config$export_sqlite_files
 imf_quarter_timestamp <- config$imf_quarter_timestamp
@@ -89,6 +90,8 @@ factset_issue_code_bridge_path <-
   file.path(factset_data_path, factset_issue_code_bridge_filename)
 factset_industry_map_bridge_path <-
   file.path(factset_data_path, factset_industry_map_bridge_filename)
+factset_manual_pacta_sector_override_path <-
+  file.path(factset_data_path, factset_manual_pacta_sector_override_filename)
 
 
 # pre-flight filepaths ---------------------------------------------------------
@@ -118,7 +121,8 @@ factset_timestamp <-
     factset_isin_to_fund_table_filename,
     factset_iss_emissions_data_filename,
     factset_issue_code_bridge_filename,
-    factset_industry_map_bridge_filename
+    factset_industry_map_bridge_filename,
+    factset_manual_pacta_sector_override_filename
   )))
 
 
@@ -135,6 +139,7 @@ stopifnot(file.exists(factset_isin_to_fund_table_path))
 stopifnot(file.exists(factset_iss_emissions_data_path))
 stopifnot(file.exists(factset_issue_code_bridge_path))
 stopifnot(file.exists(factset_industry_map_bridge_path))
+stopifnot(file.exists(factset_manual_pacta_sector_override_path))
 stopifnot(file.exists(data_prep_outputs_path))
 
 if (!update_currencies) {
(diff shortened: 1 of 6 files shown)


AlexAxthelm:
Marking as ready for review, with notes about the PRs that should be considered first.

cjyetman:
putting into draft mode because this depends on other PRs, and as such, it's nearly impossible to review properly

run_pacta_data_preparation.R
```diff
@@ -33,7 +33,6 @@ config <-
 
 asset_impact_data_path <- config$asset_impact_data_path
 factset_data_path <- config$factset_data_path
```
cjyetman, line 35:
```suggestion
factset_data_path <- config$factset_data_path
data_prep_outputs_path <- config$data_prep_outputs_path
```
required below
note: #160 may make this unnecessary in the future

run_pacta_data_preparation.R
```diff
@@ -154,31 +153,60 @@ factset_timestamp <-
...
+  factset_issue_code_bridge_path = factset_issue_code_bridge_path,
+  factset_industry_map_bridge_path = factset_industry_map_bridge_path,
+  factset_manual_pacta_sector_override_path = factset_manual_pacta_sector_override_path
+)
 
 if (!update_currencies) {
-  stopifnot(file.exists(currencies_data_path))
+  input_filepaths <- c(
+    input_filepaths,
+    currencies_data_path = currencies_data_path
```
cjyetman, line 174:
```suggestion
    currencies_preflight_data_path = currencies_preflight_data_path
```
?

cjyetman: e.g. if currencies are NOT going to be updated, then we're relying on a previously saved `currencies.rds` existing in the pre-flight path (which may or may not be the same as `data_prep_outputs_path`)

run_pacta_data_preparation.R
AlexAxthelm: #160 is relevant in the future, I agree. Discarding this suggestion, sine `data_prep_outputs_path` is defined on further down in the script since #151 was merged in.

https://github.com/RMI-PACTA/workflow.data.preparation/blob/a88994706b0f7ffc7e99fdde58548c588744c24c/run_pacta_data_preparation.R#L73-L76

run_pacta_data_preparation.R
```diff
@@ -901,14 +920,24 @@ parameters <-
...
+  path = manifest_path,
   parameters = parameters,
-  asset_impact_data_path = asset_impact_data_path,
-  factset_data_path = factset_data_path,
-  data_prep_outputs_path = data_prep_outputs_path
+  input_files = input_filepaths,
+  output_files = output_files
 )
-
+output_files <- c(output_files, manifest_path = manifest_path)
```
cjyetman, line 931: is this needed, or just being safe?

run_pacta_data_preparation.R
```diff
@@ -856,20 +878,7 @@ parameters <-
...
-      factset_isin_to_fund_table_path = factset_isin_to_fund_table_path,
-      factset_iss_emissions_data_path = factset_iss_emissions_data_path,
-      factset_issue_code_bridge_path = factset_issue_code_bridge_path,
-      factset_industry_map_bridge_path = factset_industry_map_bridge_path,
-      factset_manual_pacta_sector_override_path = factset_manual_pacta_sector_override_path
-    ),
+    input_filepaths = as.list(input_filepaths),
     preflight_filepaths = list(
       currencies_data_path = currencies_data_path
     ),
```
cjyetman, line 885:
```suggestion
```

let's remove this because the pre-flight files are now captured in `input_filepaths`

cjyetman (changes requested):
see suggestions (and line 968)

cjyetman (approved):

Issue #148, opened 2024-02-18
Export copies of scraped data


AlexAxthelm:
Include a copy of scraped data with other data files, so that data prep can be reconstructed later if neccesary.

* Currencies `currencies.rds`
* index_regions `index_regions.rds`

AlexAxthelm:
relates to #147, which needs to be updated to account for this.

cjyetman:
superseded by #158

PR #150, opened 2024-02-18
feat(app): #148 Export data from scraped files

AlexAxthelm:
Changes in this PR export data from scraped sources to file, so that all sources for a dataset are captured, and the dataset can be recreated later.

By default, exports to data prep outputs path, but can be specified with `preflight_data_path`. Writing to the outputs path by default, since this matches current behavior, and we do not have a unified inputs path that we are currently using.

Closes: #148

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -2,6 +2,7 @@ default:
   data_prep_outputs_path: "/outputs"
   asset_impact_data_path: "/inputs"
   factset_data_path: "/inputs"
+  preflight_data_path: ""
   masterdata_ownership_filename: ""
   masterdata_debt_filename: ""
   ar_company_id__factset_entity_id_filename: ""
@@ -127,3 +128,4 @@ desktop:
   data_prep_outputs_path: "./outputs"
   asset_impact_data_path: "./ai_inputs"
   factset_data_path: "./factset_inputs"
+  preflight_data_path: ""
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -96,8 +96,15 @@ factset_manual_pacta_sector_override_path <-
 
 # pre-flight filepaths ---------------------------------------------------------
 
+preflight_data_path <- config$preflight_data_path
+if (preflight_data_path == "") {
+  preflight_data_path <- data_prep_outputs_path
+}
+
+currencies_preflight_data_path <- file.path(preflight_data_path, "currencies.rds")
 currencies_data_path <- file.path(data_prep_outputs_path, "currencies.rds")
 
+index_regions_preflight_data_path <- file.path(preflight_data_path, "index_regions.rds")
 
 # computed options -------------------------------------------------------------
 
@@ -156,10 +163,16 @@ if (update_currencies) {
   currencies <- pacta.data.scraping::get_currency_exchange_rates(
     quarter = imf_quarter_timestamp
   )
+  saveRDS(currencies, currencies_preflight_data_path)
+} else {
+  logger::log_info("Using pre-existing currency data.")
+  # This requires the preflight path to be defined in the config
+  currencies <- readRDS(currencies_preflight_data_path)
 }
 
 logger::log_info("Scraping index regions.")
 index_regions <- pacta.data.scraping::get_index_regions()
+saveRDS(index_regions, index_regions_preflight_data_path)
 
 logger::log_info("Fetching pre-flight data done.")
 


jdhoffa (approved):
Seems reasonable?

cjyetman:
superseded by #158

cjyetman:
please adjust the PR title to reflect its purpose

AlexAxthelm:
@cjyetman I believe this is ready for review as-is. If you have specific changes you would like to either the description or title, let me know what they should read as.

cjyetman:
I don't understand the goal of this PR as-is, and I'm not in favor of adding a new output directory, one that was previously removed for simplification.

AlexAxthelm:
https://github.com/RMI-PACTA/workflow.data.preparation/blob/94fe89f6c48233ce7a1ca6f4c7bc0ddfd1cb27b5/run_pacta_data_preparation.R#L100-L102

By default, this does not add any new directories to the structure. It uses the existing output directory (unless overriden in the config), but exports `currencies.rds`and the `index_regions.rds` to enhance reproducibility.

If you have a different default path you would like to use, I'm open to suggestions, but this one minimizes the changes we need to make for other parts of the system.

cjyetman:
I happy with the default path, I'm not in favor of adding a new optional path which then requires continued support and consideration, especially without any reasonable justification for it.

AlexAxthelm:
The idea for that is that when running locally, or during development, or during a reconstruction of a dataset, you can specify that path to use the downloaded files (wherever they reside on the host machine). Without that, there's no good way to specify what path to use for those files.

cjyetman:
Yeah, that's why I'm not thrilled with the idea of having timestamped output directories. My gut feeling would be to go back to saving it to the inputs directory and then also copy it over to the outputs directory, which was just changed in #127. I was preemptively avoiding a potential problem with the inputs directory being read-only with that, but I don't know if that's really a concern. If not, probably best to just revert #127.

AlexAxthelm:
I think the big difference between this and reverting #127 is that here we're only assuming that the outputs path is writable, and we can preserve that behavior, while still allowing for the split between the FactSet and AI data.

cjyetman:
> I think the big difference between this and reverting #127 is that here we're only assuming that the outputs path is writable, and we can preserve that behavior, while still allowing for the split between the FactSet and AI data.

Is this a plausible concern, that the inputs directory might be read-only on Azure? I only imagined that possibility, but not sure if it's real. If so, then yeah, we're going to have to add a new directory 🤔

AlexAxthelm:
> Is this a plausible concern, that the inputs directory might be read-only on Azure?

Not just a plausible concern, but a definite fact.

I'd rather avoid having a directory there with just the scraped inputs though. My main goal with this PR is to have a way to say "from this set of inputs, do we get the same set of outputs." Not "If we lose all the outputs, can we recreate them from scratch.

cjyetman:
ok, so I guess we have to do this then... can we make the setting of the `preflight_path` conditional on `update_currencies` since they're interdependent? I'm concerned about having a config with `update_currencies = TRUE` and `preflight_path = "/some/path"` and then there being some unexpected behavior

to understand the logic...

1. if `update_currencies = TRUE` and `preflight_path = "/some/path"` - currencies are scraped and saved to specified `preflight_path` and then copied to specified `data_prep_outputs_path`
2. if `update_currencies = TRUE` and `preflight_path = ""` - currencies are scraped and saved to specified `data_prep_outputs_path` only
3. if `update_currencies = FALSE` and `preflight_path = "/some/path"` - currencies are *not* scraped but read in from "`preflight_path`/currencies.rds" and copied to specified `data_prep_outputs_path`
4. if `update_currencies = FALSE` and `preflight_path = ""` - currencies are *not* scraped but read in from "`preflight_path`/currencies.rds" and saved to specified `data_prep_outputs_path` only

Does that sound right?

AlexAxthelm:
> ok, so I guess we have to do this then... can we make the setting of the `preflight_path` conditional on `update_currencies` since they're interdependent? I'm concerned about having a config with `update_currencies = TRUE` and `preflight_path = "/some/path"` and then there being some unexpected behavior
> 
> to understand the logic...
> 
> 1. if `update_currencies = TRUE` and `preflight_path = "/some/path"` - currencies are scraped and saved to specified `preflight_path` and then copied to specified `data_prep_outputs_path`
> 2. if `update_currencies = TRUE` and `preflight_path = ""` - currencies are scraped and saved to specified `data_prep_outputs_path` only
> 3. if `update_currencies = FALSE` and `preflight_path = "/some/path"` - currencies are _not_ scraped but read in from "`preflight_path`/currencies.rds" and copied to specified `data_prep_outputs_path`
> 4. if `update_currencies = FALSE` and `preflight_path = ""` - currencies are _not_ scraped but read in from "`preflight_path`/currencies.rds" and saved to specified `data_prep_outputs_path` only
> 
> Does that sound right?

Mostly.

4. if update_currencies = FALSE and preflight_path = "" - currencies are not scraped but read in from "**`data_prep_outputs_path`**/currencies.rds" **which will (probably) cause an error on read, since the file (probably) doesn't exist.

I think a lot of your concerns about interdependency are addressed in #149, where all the input and preflight filepaths are checked for existence (if needed) prior to moving forwards.

cjyetman:
fine, I prefer to have relevant checks included where they became necessary, but these are edge cases and we'll get to it soon

cjyetman (approved):

Issue #151, opened 2024-02-18
Define unique output directory for each output run


AlexAxthelm:
Right now, exported files are all put into the `data_prep_outputs` directory, with not further organization.

The script should create a subdirectory that isolates the outputs from each run, so files are not overwritten with subsequent runs, and can be clearly identified.

Issue #152, opened 2024-02-18
Export archives of input and outputs


AlexAxthelm:
In conversation with @cjyetman and @jdhoffa , we want to export archives of the input files and output files from this process, to live alongside the "loose" files.

Having an archive provides both a clear packaging mechanism to ensure that files that belong together will stay together even through reorganizations, and also a convenient mechanism for distribution of a complete dataset.

PR #153, opened 2024-02-18
feat(app): #151 Put outputs into unique timestamped directory

AlexAxthelm:
create a unique directory for outputs, and warn if it already exists

Directory name includes both pacta timestamp (2022Q4), and creation date (approximately the time the process starts, within a few seconds if run as script).

Example: `/outputs/2022Q4_20240218T165613Z` rather than just `/outputs`

~Additionally, exports 2 `zip` archives: one of all the input files, and another of the contents of the outputs directory~ See #159 

Closes: #151

See Also: #152

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -15,7 +15,6 @@ default:
   factset_industry_map_bridge_filename: ""
   factset_manual_pacta_sector_override_filename: ""
   update_currencies: TRUE
-  timestamp_export_directory: TRUE
   export_sqlite_files: TRUE
   imf_quarter_timestamp: "2021-Q4"
   pacta_financial_timestamp: "2021Q4"
--- a/config.yml
+++ b/config.yml
@@ -15,6 +15,7 @@ default:
   factset_industry_map_bridge_filename: ""
   factset_manual_pacta_sector_override_filename: ""
   update_currencies: TRUE
+  timestamp_export_directory: TRUE
   export_sqlite_files: TRUE
   imf_quarter_timestamp: "2021-Q4"
   pacta_financial_timestamp: "2021Q4"
@@ -127,3 +128,4 @@ desktop:
   data_prep_outputs_path: "./outputs"
   asset_impact_data_path: "./ai_inputs"
   factset_data_path: "./factset_inputs"
+  timestamp_export_directory: FALSE
--- a/config.yml
+++ b/config.yml
@@ -127,4 +127,3 @@ desktop:
   data_prep_outputs_path: "./outputs"
   asset_impact_data_path: "./ai_inputs"
   factset_data_path: "./factset_inputs"
-  timestamp_export_directory: FALSE
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -33,7 +33,6 @@ config <-
 
 asset_impact_data_path <- config$asset_impact_data_path
 factset_data_path <- config$factset_data_path
-data_prep_outputs_path <- config$data_prep_outputs_path
 masterdata_ownership_filename <- config$masterdata_ownership_filename
 masterdata_debt_filename <- config$masterdata_debt_filename
 ar_company_id__factset_entity_id_filename <- config$ar_company_id__factset_entity_id_filename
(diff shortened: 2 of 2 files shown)


cjyetman:
1. exporting zip archives and adding a timestamp to the directory name should be done in separate PRs
2. I don't believe adding a timestamp to the directory name is desirable. Ideally, if a custom directory name is desired, it should be specified in the config file (with or without a timestamp). If this needs to be done in `[run_pacta_data_preparation.R](https://github.com/RMI-PACTA/workflow.data.preparation/blob/main/run_pacta_data_preparation.R), it should be done with an optional parameter to turn it on or off

AlexAxthelm:
> 1. exporting zip archives and adding a timestamp to the directory name should be done in separate PRs

I opened #159 for archive export.

> 2. I don't believe adding a timestamp to the directory name is desirable. Ideally, if a custom directory name is desired, it should be specified in the config file (with or without a timestamp). If this needs to be done in [run_pacta_data_preparation.R](https://github.com/RMI-PACTA/workflow.data.preparation/blob/main/run_pacta_data_preparation.R), it should be done with an optional parameter to turn it on or off

I've added a parameter in the config to disable this behavior, but set the default behavior to be TRUE.

This means that the default behavior is not not overwrite files/directories, and overwriting must be specifically enabled.

cjyetman:
Sorry, now that I'm thinking about this more, this is a deceptively small but significant conceptual change to what `data_prep_outputs_path` is (an output directory versus a directory where an output directory is created), and I think it's too much potential for trouble to try to support both. Let's just make this mandatory and I'll adapt my behavior when working locally.

config.yml
```diff
@@ -15,6 +15,7 @@ default:
   factset_industry_map_bridge_filename: ""
   factset_manual_pacta_sector_override_filename: ""
   update_currencies: TRUE
+  timestamp_export_directory: TRUE
```
cjyetman, line 18:
```suggestion
```

config.yml
```diff
@@ -127,3 +127,4 @@ desktop:
   data_prep_outputs_path: "./outputs"
   asset_impact_data_path: "./ai_inputs"
   factset_data_path: "./factset_inputs"
+  timestamp_export_directory: FALSE
```
cjyetman, line 130:
```suggestion
```

run_pacta_data_preparation.R
```diff
@@ -33,7 +33,6 @@ config <-
 
 asset_impact_data_path <- config$asset_impact_data_path
 factset_data_path <- config$factset_data_path
```
cjyetman, line 35:
```suggestion
factset_data_path <- config$factset_data_path
data_prep_outputs_path <- config$data_prep_outputs_path
```

run_pacta_data_preparation.R
```diff
@@ -64,6 +64,28 @@ scenario_geographies_list <- config$scenario_geographies_list
...
+    format = "%Y%m%dT%H%M%SZ",
+    tz = "UTC"
+  )
+  data_prep_outputs_path <- file.path(
+    config$data_prep_outputs_path,
+    paste(pacta_financial_timestamp, system_timestamp, sep = "_")
+  )
+} else {
+  data_prep_outputs_path <- config$data_prep_outputs_path
+}
```
cjyetman, line 80:
```suggestion
# create timestamped output directory
system_timestamp <- format(
  Sys.time(),
  format = "%Y%m%dT%H%M%SZ",
  tz = "UTC"
)
data_prep_outputs_path <- file.path(
  config$data_prep_outputs_path,
  paste(pacta_financial_timestamp, system_timestamp, sep = "_")
)
```

run_pacta_data_preparation.R
```diff
@@ -64,6 +64,24 @@ scenario_geographies_list <- config$scenario_geographies_list
...
 
+# create timestamped output directory
+system_timestamp <- format(
+  Sys.time(),
+  format = "%Y%m%dT%H%M%SZ",
+  tz = "UTC"
+)
+data_prep_outputs_path <- file.path(
+  data_prep_outputs_path,
+  paste(pacta_financial_timestamp, system_timestamp, sep = "_")
```
cjyetman, line 75: should the directory include any suffix indicating what it is, e.g. `*_dataprep_outputs`?

cjyetman:
This will break the idea of not updating the currencies because you'll never have the currencies file in the right place. Maybe it's time to put that to bed and just make updating currencies mandatory?

run_pacta_data_preparation.R
AlexAxthelm: I think having the inputs archive called out with an `_inputs` suffix (in #159) differentiates these, combined with the fact that our Azure storage puts these all in `workflow-data-prep-outputs`. If you want more differentiation though, I'm okay with that.

AlexAxthelm:
> This will break the idea of not updating the currencies because you'll never have the currencies file in the right place. Maybe it's time to put that to bed and just make updating currencies mandatory?

@cjyetman I think that might be a better topic for comment #159, but there I'm specifying an optional preflights path, which you can read an old `currencies.rds` from.

run_pacta_data_preparation.R
cjyetman: nope that's fine... just checking that it wasn't an oversight

PR #154, opened 2024-02-18
refactor(deploy): Harmonize docker mount points

AlexAxthelm:
Change from /outputs to /mnt/outputs, to match structure defined in #142 and #145

No associated issue.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,5 +1,5 @@
 default:
-  data_prep_outputs_path: "/outputs"
+  data_prep_outputs_path: "/mnt/outputs"
   asset_impact_data_path: "/inputs"
   factset_data_path: "/inputs"
   masterdata_ownership_filename: ""
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,4 +10,4 @@ services:
         target: /inputs
       - type: bind
         source: ${HOST_OUTPUTS_PATH}
-        target: /outputs
+        target: /mnt/outputs


cjyetman:
Gonna take me some time to test this locally to make sure it works in the local Docker context. Some quick thoughts though...

1. I don't see any obvious logical explanation for why the outputs path is under `/mnt` while the inputs directories are not.
2. I will be hyper-critical of any changes to `config.yml`, and in particular, in order of decreasing severity, changes to the "default" config and changes to the default timestamp configs, e.g. "2022Q4"
3. I'm leaning towards not defining paths in any of the default configs, since the paths are always context specific, so in the near future this will probably look more like...
```yml
default:
  data_prep_outputs_path: ""
  asset_impact_data_path: ""
  factset_data_path: ""

2022Q4_azure:
  inherits: 2022Q4
  data_prep_outputs_path: "/mnt/outputs"
  asset_impact_data_path: "/inputs"
  factset_data_path: "/inputs"

2022Q4_docker:
  inherits: 2022Q4
  data_prep_outputs_path: "/outputs"
  asset_impact_data_path: "/inputs"
  factset_data_path: "/inputs"

2022Q4_desktop:
  inherits: 2022Q4
  data_prep_outputs_path: "/path/to/local/outputs"
  asset_impact_data_path: "/path/to/local/inputs"
  factset_data_path: "/path/to/local/inputs"
```

cjyetman:
prefer to wait until #161 merges and add this new context specific path to the new "docker" config instead

cjyetman (approved):
tested, seems to work... still uncertain about the logic of outputs living under `/mnt` but inputs not

AlexAxthelm:
This was compensating for the fact that I forgot to move the outputs directory in #145

Issue #138, opened 2024-02-16
consider copying `.git` into Docker container to facilitate precise tracking of code used to generate outputs in manifest


cjyetman:
I think by not including the `.git` directory in #129, the manifest has now lost the functionality of tracking precisely what version of the run script was used, and even if any uncommitted local changes had been made (e.g. to the config file or the script itself) when running in a Docker container.

We now see something like this...
```json
  "git": {
    "local_git_tag": "not a git repo",
    "local_git_hash": "not a git repo",
    "local_git_status": "not a git repo"
  },
```

as opposed to what we used to see and/or what we can see if running the "desktop"/local-RStudio way, like this...
```json
  "git": {
    "local_git_tag": "no git tags found",
    "local_git_hash": "f97e472a5d03c8c1c95ebfa3c51b7e2534fc892e",
    "local_git_status": [
      {
        "status": "",
        "filename": "M config.yml"
      }
    ]
  },
```

@AlexAxthelm or @jdhoffa do you have opinions to the pros-cons of including the `.git` directory when copying over to the Docker container? I'm in inclined to start copying it in again because I think this git status information is very useful for tracking how the data.prep outputs were generated (I don't think there is any other indication of, for instance, what version of [run_pacta_data_preparation.R](https://github.com/RMI-PACTA/workflow.data.preparation/blob/main/run_pacta_data_preparation.R) was used). Or any other ideas on how to capture that info in the Docker container?

AlexAxthelm:
If we copy just the `.git/HEAD` file, and `/git/refs/heads` directory, then we can get the current commit (but not clean/dirty status) without bringing in the entire history.

`HEAD` is a oneline file that has the name of the ref, `/refs/heads/*` are files with the commit SHAs of the HEAD of each ref.

So from my current copy of this repo, on `main` (which happens to be at f14e5af7c8b8e03fc7c8739dc1f1cc09fe3dc95c as of right now):

```
sh-3.2$ pwd
/Users/aaxthelm/Documents/pacta/workflow.data.preparation

sh-3.2$ git rev-parse HEAD
f14e5af7c8b8e03fc7c8739dc1f1cc09fe3dc95c

sh-3.2$ git branch
  azure-vm
  feature/azure-container
  feature/logger
* main
  remove-fact-set-database-code

sh-3.2$ cat .git/HEAD
ref: refs/heads/main

sh-3.2$ cat .git/refs/heads/main
[...]

PR #155, opened 2024-02-18
build(docker): #138 Allow .git in build context

AlexAxthelm:
We may want to have a more elegant solution in the future (see discussion in #138), but this works for now to include git status in the manifest export.

Closes: #138

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 *
 # include certain files
 !.env
+!.git
 !config.yml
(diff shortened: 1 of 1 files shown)


AlexAxthelm:
Suggested action: merge this PR for now, since impact on build process is small, and flag a new ticket to optimize based on the discussions in #138.

cjyetman (approved):
#157 opened to track looking for a leaner method of capturing git status

PR #158, opened 2024-02-22
save `index_regions` to outputs

### LINKED ISSUE #148: Export copies of scraped data

cjyetman:
supersedes #150 
closes #148


jdhoffa (approved):
seems reasonable!

PR #159, opened 2024-02-22
feat(app): #152 Add functionality to export archives

AlexAxthelm:
Adds config option to export archives of the data inputs and output directories, at same level of output directory (no nesting)

Closes: #152
Relates to #153

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -16,6 +16,7 @@ default:
   factset_manual_pacta_sector_override_filename: ""  # a single filename of the intended factset_manual_pacta_sector_override RDS (must exist in the `factset_data_path` directory)
   update_currencies: TRUE  # a single `TRUE` or `FALSE` specifying whether currency data should be pulled on-the-fly (or a previously pulled version should be used)
   export_sqlite_files: TRUE  # a single `TRUE` or `FALSE` specifying whether sqlite versions of some output files should be included in the outputs
+  export_archives: TRUE # Boolean: Export zip archives of input files and output files?
   imf_quarter_timestamp: ""  # a single string specifying the IMF quarter timestamp (for pulling of currency data) in the form "YYYY-QX", e.g. "2021-Q4"
   pacta_financial_timestamp: ""  # a single string specifying the PACTA quarter timestamp in the form "YYYYQX", e.g. "2021Q4"
   market_share_target_reference_year: 0  # a single integer specifying the market share target reference year in the form YYYY, e.g. 2021
--- a/config.yml
+++ b/config.yml
@@ -16,6 +16,7 @@ default:
   factset_manual_pacta_sector_override_filename: ""
   update_currencies: TRUE
   export_sqlite_files: TRUE
+  export_archives: TRUE
   imf_quarter_timestamp: "2021-Q4"
   pacta_financial_timestamp: "2021Q4"
   market_share_target_reference_year: 2021
@@ -127,3 +128,4 @@ desktop:
   data_prep_outputs_path: "./outputs"
   asset_impact_data_path: "./ai_inputs"
   factset_data_path: "./factset_inputs"
+  export_archives: FALSE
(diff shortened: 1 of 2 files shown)


config.yml
```diff
@@ -127,3 +128,4 @@ desktop:
   data_prep_outputs_path: "./outputs"
   asset_impact_data_path: "./ai_inputs"
   factset_data_path: "./factset_inputs"
+  export_archives: FALSE
```
cjyetman, line 131:
```suggestion
```
While I appreciate trying to maintain "desktop" behavior, I'd like it to remain as close to "default" as reasonable.

run_pacta_data_preparation.R
```diff
@@ -890,6 +890,38 @@ for (pkg_name in pacta_packages) {
   )
 }
 
+# Create archive files
+if (config[["export_archives"]]) {
```
cjyetman, line 894:
```suggestion
if (export_archives) {
```
Generally speaking, I currently prefer this approach, but to remain in-sync with how other parameters are currently being used/managed in this script, I'd rather see a line like `export_archives <- config$export_archives` above, e.g. between [these two lines](https://github.com/RMI-PACTA/workflow.data.preparation/blob/e27b2d443a6322874d4883187275cd41bf6b30cd/run_pacta_data_preparation.R#L50C1-L51), and maybe in the future we convert everything to the strategy of using parameters directly from the `config` object/list.

cjyetman: tracking switch to new method in #160

run_pacta_data_preparation.R
```diff
@@ -890,6 +890,38 @@ for (pkg_name in pacta_packages) {
...
+  logger::log_info("Exporting input and output archives.")
+
+  logger::log_debug("Creating inputs zip file.")
+  inputs_zip_file_path <- paste0(data_prep_outputs_path, "_inputs.zip")
+  logger::log_trace("Zip file path: \"{inputs_zip_file_path}\".")
+  zip(
+    zipfile = inputs_zip_file_path,
+    files = unlist(parameters[["input_filepaths"]]),
```
cjyetman, line 902:
```suggestion
    files = normalizePath(unlist(parameters[["input_filepaths"]])),
```

zip will fail if these paths are not absolute and canonical

run_pacta_data_preparation.R
```diff
@@ -890,6 +891,38 @@ for (pkg_name in pacta_packages) {
...
+    extras = c(
+      "--junk-paths", # do not preserve paths
+      "--no-dir-entries", # do not include directory entries
+      "--quiet" # do not print progress to stdout
+    )
+  )
+  logger::log_debug("Inputs archive created.")
+
+  logger::log_debug("Creating outputs zip file.")
+  outputs_zip_file_path <- paste0(data_prep_outputs_path, ".zip")
```
cjyetman, line 913: Is this dangerous? Effectively, this will drop a zip file in the parent directory of the specified output directory, which could be inaccessible. Same with the `inputs_zip_file_path`

cjyetman: this is less of a problem with latest changes to #153 to make the timestamped output directory non-optional?

AlexAxthelm: In general, this is intended to be used in conjunction with the changes in #153. I split it out into a separate PR on your request. 

But in general, if someone if using a generic `./outputs` and `./inputs` (not timestamped), then exporting the archive is probably a bad idea. But since I've made the default in #153 to use time unique directories, that dangerous behavior is explicitly opt-in.

AlexAxthelm: The only time I would expect this to path to be unwritable is when the user is writing directly to the directory which is volume mounted to the docker container. That's pretty far off-label usage so I'm not particularly worried about it, but if you want to flag an issue to do a warning check, I can add it in later.

cjyetman: with #153 merged and it forcing outputs into a subdirectory, and therefore the zip landing in the originally specified `data_prep_outputs_path`, I think this is not a concern anymore

cjyetman (approved):

Issue #160, opened 2024-02-22
consider using imported parameters directly from the `config` object, rather than copying them to global environment


cjyetman:
Currently, parameters imported from the [`config.yml`](https://github.com/RMI-PACTA/workflow.data.preparation/blob/main/config.yml) are copied into global variables one-by-one [here](https://github.com/RMI-PACTA/workflow.data.preparation/blob/e27b2d443a6322874d4883187275cd41bf6b30cd/run_pacta_data_preparation.R#L34-L65), and then used via these global variables. It may be preferable to access the parameters directly from the `config` object rather than creating numerous global variables. e.g.

change this...
```r
data_prep_outputs_path <- config$data_prep_outputs_path

...

currencies_data_path <- file.path(data_prep_outputs_path, "currencies.rds")
```

to this...
```r
currencies_data_path <- file.path(config[["data_prep_outputs_path"]], "currencies.rds")
```

AlexAxthelm:
I'm all for this. Noting that where we currently do the conversion from config elements to top-level R objects (`x <- config$x`) should be replaced with a check that the elements we are using exist in the config, as a way to preserve fail-fast behavior

```r
expected_keys <- c(
  "asset_impact_data_path",
  "factset_data_path",
  "data_prep_outputs_path",
  ...
)
stopifnot(all(expected_keys %in% names(config)))
```

PR #161, opened 2024-02-23
cleanup config

cjyetman:
- removes timestamp specific values from "default" config
- moves context specific configs to set directory paths towards the top
- makes timestamp configs inherit directory paths from new "docker" config

note: #145 will "officialize" the 2022Q4 values

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -33,9 +33,9 @@ default:
 
 desktop:
   inherits: 2022Q4
-  data_prep_outputs_path: "~/Desktop/dataprep_local/outputs"
-  asset_impact_data_path: "~/Desktop/dataprep_local/inputs"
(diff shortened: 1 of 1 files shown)


jdhoffa (approved):

AlexAxthelm:
For future reference, I think it would have been better to remove the "empty" keys entirely, especially for paths. One good way to do this while still specifying a complete spec would be to comment out those lines.

As it stands, the default behavior will cause errors and incorrect expectations. files will attempt to write to incorrect directories and the other parameters will cause unexpected behaviors in the resulting dataset. It would be better to have those explicitly error out when we attempt to use those keys as part of assignment to variables (possibly with a check on NULL)

relates to #160, as doing a check on `expected_keys %in% names(config)` would be a good way to do a check for completeness while failing fast.

PR #167, opened 2024-02-25
add `config.schema.json`

cjyetman:


cjyetman:
@AlexAxthelm this is still in draft mode, assuming you may have some suggestions. Also still need to figure out appropriate values for the "$id", "title", and "description" keys... and configure the GH Action appropriately and make sure it works as expected.

AlexAxthelm:
Exciting! I'll take a closer look later!

AlexAxthelm:
@cjyetman Noting a couple of things here from the docs used in the action:

First: it doesn't look like we're actually validating against the schema you wrote. From [the docs](https://github.com/GrantBirki/json-yaml-validate?tab=readme-ov-file#inputs-):

> Input | Required? | Default | Description
> -- | -- | -- | --
> yaml_schema | false | "" | The full path to the YAML schema file (e.g. ./schemas/schema.yaml) - Default is "" which doesn't enforce a strict schema

So right now it's just doing a check on "is this a valid yaml file", not "does this yaml file match this schema".

Second: I don't think we can make a good schema that will cover the inheritance that we're counting on with `config::get()`. 

The Schema should model a _fully-defined_ object, but as we have the files written (`2023Q4_azure` inherits from `2023Q4` inherits from `docker` inherits from `default`), none of those objects will necessarily have all the correct keys.

Here's an article about modeling inheritance with JSON schema: https://json-schema.org/blog/posts/modelling-inheritance but the short story is that it isn't as intuitive as `config` makes it.

I think this absolutely makes sense for us to chase down, but it might involve bigger changes to how we handle/import configs than just doing the check.

AlexAxthelm:
One "workaround" option would be that for each of the configs, we read them in with `config::get()` (to deal with inheritance), and write them to file before checking them against schema.

cjyetman:
Yeah, I had a feeling this validator action isn't going to work with the inheritance easily. I'm definitely not wedded to it, just thought it would be cool if it worked.

cjyetman:
gonna remove the validator action because I don't think there's any reasonable way of making it work as expected/desired

cjyetman:
While this was a cool experiment, I don't know if this would ever get used and it has already fallen out of date, which suggests to me this will be more of a maintenance burden than useful, so I'm gonna close it.

PR #168, opened 2024-02-26
use params from `config`, not global env

### LINKED ISSUE #160: consider using imported parameters directly from the `config` object, rather than copying them to global environment

cjyetman:
closes #160 

Use the parameters set in the config file directly from the `config` object rather than copying each one to a new variable in the global environment (and check that they necessary keys exist when the config is imported).

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -24,12 +24,14 @@ if (Sys.getenv("R_CONFIG_ACTIVE") != "desktop") {
 }
 
 config_name <- Sys.getenv("R_CONFIG_ACTIVE")
-config <-
+raw_config <-
   config::get(
     file = "config.yml",
     config = config_name,
     use_parent = FALSE
   )
+
+config <- raw_config
 
 expected_keys <- c(
   "asset_impact_data_path",
@@ -880,6 +882,7 @@ package_news <-
 parameters <-
   list(
     config_name = config_name,
+    raw_config = unclass(raw_config),
     config = unclass(config),
     input_filepaths = as.list(input_filepaths),
     timestamps = list(


jdhoffa (dismissed):

cjyetman:
@AlexAxthelm understand if you want to pause on this one to avoid massive merge conflicts with other open PRs 🤣

run_pacta_data_preparation.R
```diff
@@ -31,124 +31,127 @@ config <-
...
 # create timestamped output directory
 system_timestamp <- format(
   Sys.time(),
   format = "%Y%m%dT%H%M%SZ",
   tz = "UTC"
 )
-data_prep_outputs_path <- file.path(
-  data_prep_outputs_path,
-  paste(pacta_financial_timestamp, system_timestamp, sep = "_")
+config[["data_prep_outputs_path"]] <- file.path(
```
AlexAxthelm, line 77: Since we're altering the values in `config` here, I think it would be prudent to capture the raw config, and also add that to `manifest.json`

```r
raw_config <- config::get(...)
config <- raw_config

#...
parameters <-
  list(
    config_name = config_name,
    config = unclass(config),
    raw_config = unclass(raw_config),
    input_filepaths = as.list(input_filepaths),

```

AlexAxthelm (dismissed):
Approved with one suggested change (can be flagged in issue for later resolution)

run_pacta_data_preparation.R
cjyetman: done

AlexAxthelm (approved):
Awesome!

cjyetman:
@AlexAxthelm should I wait to merge to avoid merge conflicts on your other PRs?

AlexAxthelm:
@cjyetman I think we can merge this, since the other PRs that are intended to be merged don't touch the R file.

PR #170, opened 2024-02-26
include `ent_entity_affiliates_last_update` in manifest

cjyetman:
This was removed in #73 when the FactSet database pulling info was removed, but `ent_entity_affiliates_last_update` is still created above and its value/info is still worthwhile to have in the manifest.


jdhoffa (approved):
🚀

AlexAxthelm (approved):

PR #175, opened 2024-02-26
copy only needed files to `/workflow.data.preparation/` in Docker image

### LINKED ISSUE #173: Explicitly COPY only needed files in Docker build
See: https://github.com/RMI-PACTA/pacta.data.preparation/pull/342#issuecomment-1964149159

As only a few files from this repo are needed in the docker image to run, consider only COPYing those files.

Would also allow for a more permissive `.dockerignore`

Relates to #129

cjyetman:
closes #173 

- moves `WORKDIR /workflow.data.preparation` up so we don't have to keep specifying it
- copies `.env` and `DESCRIPTION` simultaneously
- copies `.git/` separately because I can't figure out if/how to do a multi-copy including a sub-directory
- copies `config.yml` `run_pacta_data_preparation.R` simultaneously

should we also remove the explicit `.dockerignore` @AlexAxthelm?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,14 +9,15 @@ RUN Rscript -e '\
   )) \
   '
 
-COPY .env /.env
-COPY DESCRIPTION /DESCRIPTION
+WORKDIR /workflow.data.preparation
+COPY .env DESCRIPTION ./
 
 RUN Rscript -e '\
   readRenviron(".env"); \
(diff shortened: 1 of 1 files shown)


AlexAxthelm (approved):
I generally prefer specifying the absolute paths in COPY, but this works fine.

PR #183, opened 2024-03-01
Ci/increase timeout for imf

AlexAxthelm:
Increase server timeout to deal with server unavailability.

Suggestion from @cjyetman on Teams:

> this was happening to me yesterday. I had to bump the timeout to 600s, but eventually it started responding faster again (after 10 min or so)


jdhoffa (approved):

run_pacta_data_preparation.R
```diff
@@ -204,7 +204,8 @@ if (config[["update_currencies"]]) {
     currencies_preflight_data_path = currencies_preflight_data_path
   )
   currencies <- pacta.data.scraping::get_currency_exchange_rates(
-    quarter = config[["imf_quarter_timestamp"]]
+    quarter = config[["imf_quarter_timestamp"]],
+    max_seconds = 600L
```
cjyetman, line 208: is 10 minutes a reasonable amount of waiting time to bake in to the code? honestly, not sure what makes sense / what you're comfortable with. I'm ok with it if you're ok with it

cjyetman (approved):

run_pacta_data_preparation.R
AlexAxthelm: I mean, most of the time it should be ~0. I don't think that under normal conditions this will have a big effect.

PR #188, opened 2024-03-10
add badge to README

### LINKED ISSUE #187: add badges to README

cjyetman:
closes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+<!-- badges: start -->
+[![Lifecycle: experimental](https://img.shields.io/badge/lifecycle-experimental-orange.svg)](https://lifecycle.r-lib.org/articles/stages.html#experimental)
+<!-- badges: end -->
+
 # workflow.data.preparation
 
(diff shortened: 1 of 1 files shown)


AlexAxthelm (approved):

README.md
```diff
@@ -1,3 +1,7 @@
+<!-- badges: start -->
+[![Lifecycle: experimental](https://img.shields.io/badge/lifecycle-experimental-orange.svg)](https://lifecycle.r-lib.org/articles/stages.html#experimental)
```
AlexAxthelm, line 2: Do we actually consider this repo to be experimental? It feels pretty stable to me (and we're using it in PROD)

jdhoffa: Relates to https://github.com/RMI-PACTA/practices/issues/6

cjyetman: 🤷🏻 it "feels" very experimental to me... I was just reading about this and trying to understand how any of the repos I maintain could possibly go from feeling experimental to any other stage

jdhoffa: Yeah the point of linking that issue, is that I don't think we have any internal definition of what these flags would even mean or how to define them self-consistently.

I would propose that as our next DevHangouts so we can get a mutual definition there?

jdhoffa: yup definitely!
In general, I think it would be good to review all of the lifecycle statuses of our repositories every quarter, and decide collectively which ones we want to bump to "stable" or supersede or deprecate?

PR #206, opened 2024-04-02
include the FactSet inputs' manifest in the outputs manifest

### LINKED ISSUE #132: consider copying `factset-export-manifest.json` to outputs and/or copying info to `manifest.json`
[`workflow.factset`](https://github.com/RMI-PACTA/workflow.factset) exports a `factset-export-manifest.json` with its standard outputs that details everything one might want to know about the FactSet data files in that export (thanks @AlexAxthelm!)

We should consider whether we should copy the `factset-export-manifest.json` file to the output directory and/or include the data in the `factset-export-manifest.json` in the `manifest.json` that [workflow.data.preparation](https://github.com/RMI-PACTA/workflow.data.preparation) adds to its output directory

⚠️ It would be important to consider whether the implementation forces the `factset-export-manifest.json` file to be a mandatory input, or if it only does this *if* the file is available.

[AB#10385](https://dev.azure.com/RMI-PACTA/d77fc8f8-abac-42ae-9e31-846f9901290d/_workitems/edit/10385)

cjyetman:
- closes #132

### DIFF
--- a/run_pacta_data_preparation.R
+++ b/run_pacta_data_preparation.R
@@ -763,6 +763,20 @@ ent_entity_affiliates_last_update <-
   pull(ent_entity_affiliates_last_update) %>%
   unique()
 
+# include FactSet inputs manifest if it is available
+factset_manifest_path <-
+  file.path(
+    config[["factset_data_path"]],
+    paste0(factset_timestamp, "-factset-export-manifest.json")
+  )
+if (file.exists(factset_manifest_path)) {
+  factset_manifest <- jsonlite::fromJSON(factset_manifest_path)
+  factset_manifest <- factset_manifest[c("files", "data_timestamp", "start_time")]
+} else {
+  logger::log_warn("FactSet manifest file was not found: \"{factset_manifest_path}\".")
+  factset_manifest <- NULL
+}
+
 # include PACTA packages NEWS.md test in the parameters to export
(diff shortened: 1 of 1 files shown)


jdhoffa (approved):

cjyetman:
will wait for @AlexAxthelm's review on this one

run_pacta_data_preparation.R
```diff
@@ -777,6 +777,20 @@ ent_entity_affiliates_last_update <-
...
 
+# include FactSet inputs manifest if it is available
+factset_manifest_path <-
+  file.path(
+    config[["factset_data_path"]],
+    paste0(factset_timestamp, "-factset-export-manifest.json")
+  )
+if (file.exists(factset_manifest_path)) {
+  factset_manifest <- jsonlite::fromJSON(factset_manifest_path)
+  factset_manifest <- factset_manifest[c("files", "data_timestamp", "start_time")]
```
AlexAxthelm, line 788:
```suggestion
  factset_manifest <- factset_manifest[c("files", "data_timestamp", "start_time", "metadata_creation_time_date")]
```
Not required, since we're effectively getting a unique timestamp from the `factset_timestamp`, but I think this might be worth bringing along as an additional point.

AlexAxthelm (dismissed):
🚢

cjyetman:
force merging because the only change was adding @AlexAxthelm's suggestion

PR #208, opened 2024-04-04
use `pacta.data.preparation::determine_relevant_years()`

cjyetman:
- towards #94 
- depends on https://github.com/RMI-PACTA/pacta.data.preparation/pull/357


jdhoffa (approved):
lgtm

PR #209, opened 2024-04-04
use `pacta.data.preparation::prepare_scenarios_long()`

cjyetman:
- towards #94
- depends on https://github.com/RMI-PACTA/pacta.data.preparation/pull/358


cjyetman:
closing this and not changing the scenario bits until we switch to using pacta.scenario.data.preparation

PR #210, opened 2024-04-04
use `pacta.data.preparation::standardize_asset_type_names()`

cjyetman:
- towards #94 
- depends on https://github.com/RMI-PACTA/pacta.data.preparation/pull/359


jdhoffa (approved):
lgtm

PR #211, opened 2024-04-04
Use `pacta.data.preparation::prepare_masterdata_debt()`

cjyetman:
- towards #94 
- depends on https://github.com/RMI-PACTA/pacta.data.preparation/pull/360


jdhoffa (approved):
lgtm

PR #213, opened 2024-04-05
use new ISS prep functions

cjyetman:
- towards #94 
- depends on https://github.com/RMI-PACTA/pacta.data.preparation/pull/361


jdhoffa (approved):
lgtm

PR #215, opened 2024-04-05
default `export_archives` to `FALSE`

cjyetman:
Since exporting the archives is very expensive (time and memory), I'm setting this default to `FALSE` and explicitly opting-in for the azure configs (where it is generally desired).


jdhoffa (approved):
lgtm, but I guess @AlexAxthelm is the more appropriate reviewer.

AlexAxthelm (approved):
I suppose this makes sense, since the only time that we actually want the archives is on the azure runs.

Issue #182, opened 2024-03-01
Move output directory creation to after pre-flight activities


AlexAxthelm:
Currently, the output directory creation step occuers near the beginning of the process. importantly, prior to the pre-flight checks and data scraping, meaning that if any of those fail, the process will create an empty directory and then error out, leaving it empty.

Suggest moving output directory creation to later in the process, just prior to when we start writing files.

https://github.com/RMI-PACTA/workflow.data.preparation/blob/819e1d1ab9144bb66fd2c000a93168340553f1bb/run_pacta_data_preparation.R#L84-L90

https://github.com/RMI-PACTA/workflow.data.preparation/blob/819e1d1ab9144bb66fd2c000a93168340553f1bb/run_pacta_data_preparation.R#L198-L224

[AB#10858](https://dev.azure.com/RMI-PACTA/d77fc8f8-abac-42ae-9e31-846f9901290d/_workitems/edit/10858)

cjyetman:
I could add on `on.exit()` hook to remove the directory if it doesn't contain any files (or on any condition), which should run even if there's an error.

PR #224, opened 2024-04-18
perf(app): #182 move directory creation later

AlexAxthelm:
Move directory creation to just prior to writing the first file to output directory, so that it is not created unnecessarily.

Closes: #182


AlexAxthelm:
From: https://github.com/RMI-PACTA/workflow.data.preparation/issues/182#issuecomment-1972878210

> I could add on `on.exit()` hook to remove the directory if it doesn't contain any files (or on any condition), which should run even if there's an error.

I think that delaying creation is a better solution than creating and then removing (robust against system crashes)

AlexAxthelm:
> Where do the pre-flight files get saved if `config[["preflight_data_path"]] == ""`?

Hmm. fair point. That would error out. I don't see a good way to resolve that, so I think we can live with the extra directories being created.

PR #225, opened 2024-04-23
config: use FactSet data pulled in April

AlexAxthelm:
Includes new ISINs, as well as update Fund data as requested from a CH user, for the PA2024CH project.


Antoine-Lalechere (approved):
lgtm

cjyetman:
Thanks. @jdhoffa is going to check with all relevant people to make sure they understand the consequences of changing the data midstream in a COP project, and I will approve/merge once I get a positive feedback from that.

cjyetman:
ack, sorry... for 2023Q4 this is fine to do now

cjyetman (approved):

AlexAxthelm:
> Thanks. @jdhoffa is going to check with all relevant people to make sure they understand the consequences of changing the data midstream in a COP project, and I will approve/merge once I get a positive feedback from that.

I think this is still relevant, since this change is specifically intended to resolve issues with PA2024CH.

jdhoffa:
@Antoine-Lalechere and @AlexAxthelm thanks for this!
The PR LGTM and I am 99% sure we are going to go this direction and merge this, but would like to hold off until we get a final decision based on all associated risks (see comment here: https://dev.azure.com/RMI-PACTA/2DegreesInvesting/_workitems/edit/10208#30762716)

jdhoffa:
> actually, I think we should consider this more... I think it might be prudent to create a new `2023Q4_PA2024CH` config because it's clear that a situation can easily occur where the config for a COP project is different than what we want for `GENERAL`. @jdhoffa can we discuss a strategy for this

CJ, to this end I have created a new issue here: https://github.com/RMI-PACTA/workflow.data.preparation/issues/226

Given that GENERAL has not yet been released, and we have no other COP projects, I am tempted to deprioritize this for now, but I think this feature will be something we should tackle if ever we intend to do another COP project this year (e.g. using 2023Q4 data)

jdhoffa:
Based on this decision: https://dev.azure.com/RMI-PACTA/2DegreesInvesting/_workitems/edit/10208#30766213

We are good to go!

@AlexAxthelm feel free to merge when ready.

jdhoffa (approved):

PR #242, opened 2024-07-22
`docker-compose` is now `docker compose`

cjyetman:


AlexAxthelm (approved):
NB: Everything works for these commands (which are the most common), but last I checked, `docker-compose` and `docker compose` are not fully interchangeable, and there were some differing behaviors in the commands (usually the less-used ones).

I might be (hope I am) wrong, but just something to be aware of.

cc @jdhoffa

cjyetman:
https://docs.docker.com/compose/migrate/#how-do-i-switch-to-compose-v2

jdhoffa:
Yeah I've found `docker compose` to be a pretty huge pain in the butt in the past, and always default to using `docker-compose` still. 

But if it's your decision to start using this new `cli` and you want to deal with whatever issues that may introduce, then happy with that!

cjyetman:
I get errors when I use `docker-compose`, so `docker compose` is the only way that currently works for me.

PR #243, opened 2024-07-22
reserve appropriate amount of RAM in docker-compose.yml

cjyetman:
Use the docker-compose file to specify a minimum amount of RAM needed to run the process to avoid problems stemming from attempting to run the process with an externally fixed amount of RAM below what is needed https://docs.docker.com/compose/compose-file/deploy/#memory


cjyetman:
doesn't seem to work as I hoped/expected, so closing

PR #244, opened 2024-11-19
Add documentation about copying pacta data for use with webapp

AlexAxthelm:


jdhoffa (approved):

cjyetman (approved):
To be honest, I don't understand what most of this means / can't verify it myself

